PR #142, opened 2024-11-28
Rework ID Capture supported documents pages

m0oz:


m0oz:
* Align supported documents with the 7.0 API (FullDocumentScanner and SingleSideScanner)
* Simplify ID Validate supported documents
* Fix the broken links in the integration tiles on https://docs.scandit.com/id-scanning/
* Make the human-readable supported documents expandable per continent to improve readability

PR #153, opened 2024-12-13
fix: SDC-23791 sdc-lib references

eatsjobs:


docs/sdks/web/add-sdk.md
```diff
@@ -222,33 +222,98 @@ import * as SDCBarcode from '@scandit/web-datacapture-barcode';
...
+```
+
+:::note
+We suggest serving the library files with the proper headers `Content-Length` and `Content-Encoding` if any compression is present.
+In case of totally missing information, we show an estimated progress.
+:::
 
 ## Additional Information
 
+### Server Side Rendering and Server Side Generation
```
eatsjobs, line 308: @gpene I see the content of this MR is not either in staging nor in production. did you miss it in some merge conflict resolution?

eatsjobs: ah my bad I was expecting it in the wrong place nevermind

PR #177, opened 2025-03-12
7.1.2 Release Notes

gpene:
+ update versions


docs/sdks/web/release-notes.md
```diff
@@ -33,7 +33,17 @@ keywords:
...
+* Fixed an issue in `SparkScanView` that could result in duplicate trigger buttons.
+
+## 7.1.2
+
+**Released**: March 13, 2025
+
+### Bug Fixes
+
+* Fixed a rare issue in SparkScan that would not allow for properly drawing the barcode location.
+* Fixed an issue in SparkScan where the floating button would appear in the center as opposed to bottom-right of the screen.
```
moritzhartmeier, line 45: @gpene this was just confirmed as not working in web (it is working on android and ios) and we will only fix it in 7.1.3.

We should also add this to all frameworks that implement SparkScan because they inherit it from Android & iOS.

PR #349, opened 2026-04-29
Add skills install callouts and rework product navigation

raffaelefarinaro:
## Summary

This PR makes two changes aimed at helping developers find and adopt the right Scandit product faster.

**1. Install-skill callout on every product/framework docs page.** A new `SkillsCallout` component appears inline on each product+framework documentation page, offering a one-click install command for the matching skill from `scandit/skills` (Any AI / Claude Code / Cursor tabs). The same callout, in its "shared" variant, sits below the product cards on the homepage so users who haven't picked a product yet can install the `data-capture-sdk` picker skill and ask their coding agent which Scandit product fits their use case. Coverage is complete: every product+framework intersection that has a skill in the repo now renders the callout, including the multi-skill .NET and MatrixScan AR iOS combinations. Install actions emit a unified `skills_command_copied` PostHog event so we can measure adoption.

**2. Homepage and sidebar reworked to match the new product marketing direction.** Products are recategorized in the left sidebar and on the homepage grid so the navigation mirrors the way the products are now positioned (Data Capture grouping, skills banner placement, refined Smart Label Capture intro). The aim is to make it obvious which product to start with and where to find it, without users having to know the legacy internal naming.

Plus smaller polish carried along: tightened tab copy ("coding agent", "integrate, debug, and customize"), softer `CommandBlock` styling in light mode, inline-code wrapping in admonitions, and updated references to the renamed `scandit/skills` repo.


src/data/skills.json
```diff
@@ -0,0 +1,45 @@
+{
+  "repo": "https://github.com/scandit/scandit-sdk-skills",
```
moritzhartmeier:
```suggestion
  "repo": "https://github.com/scandit/skills",
```
It does forward correctly currently, I assume it was renamed.

moritzhartmeier (changes requested):

src/data/skills.json
raffaelefarinaro: yes we just renamed it

src/components/HomePage/data/createLabelScanningArr.tsx
```diff
@@ -28,26 +28,22 @@ export function createLabelScanningArr(framework: string) {
...
 
-  const allCardsArray = [
+  return [
     {
       name: LabelScanning.LabelScanning,
-      text: "Capture Multiple Barcodes and Text from Labels",
+      text: "Capture expiry dates, prices, weights, VIN, IMEI, or serial numbers",
+      apis: "Smart Label Capture",
       icon: <LabelCapture />,
       isActive: frameworkData?.labelScanning?.includes(
```
lucatorella, line 36: This still fails `npm run typecheck`: `frameworkCards` is inferred as a narrow union, and some `additional` entries do not have `labelScanning`, so this access is not valid for every possible return from `findFrameworkData()`.

A small fix is to widen the source array and return type instead of relying on inference:

```ts
// frameworkCardsArr.tsx
import type { FrameworkCardType } from "../../constants/types";

export const frameworkCards: FrameworkCardType[] = [
  ...
];

// createLabelScanningArr.tsx
function findFrameworkData(): FrameworkCardType | undefined {
  return (
    frameworkCards.find((item) => item.framework === framework) ??
    frameworkCards
      .flatMap((item) => item.additional ?? [])
      .find((item) => item.framework === framework)
  );
}
```

I would also make `isActive` default to `false` with `?? false`, so the returned card shape stays stable.

docs/label-definitions.mdx
```diff
@@ -0,0 +1,9 @@
+---
+id: label-definitions
+title: Supported Labels
+sidebar_label: Supported Labels
+---
+
+import {Redirect} from '@docusaurus/router';
+
+<Redirect to="/sdks/web/label-capture/label-definitions/" />
```
lucatorella, line 9: Because this page is only a client-side redirect, it will be picked up by the LLM docs generator unless it is added to `llmsRootRedirectOnlyDocs`. That makes `llms.txt` / `llms-full.txt` include a useless `Supported Labels` entry whose content is just the redirect component instead of the canonical Web label-definition docs.

Suggested fix in `docusaurus.config.ts`:

```ts
const llmsRootRedirectOnlyDocs: string[] = [
  "docs/barcode-scanning.mdx",
  ...
  "docs/id-scanning.mdx",
  "docs/label-definitions.mdx",
  "docs/migrate-5-to-6.mdx",
  ...
];
```

src/components/HomePage/data/createLabelScanningArr.tsx
raffaelefarinaro: Done — typed `frameworkCards` as `FrameworkCardType[]` in `frameworkCardsArr.tsx`, simplified `findFrameworkData` to your suggested form, and added `?? false` on `isActive` so the card shape stays stable. `npm run typecheck` now passes.

docs/label-definitions.mdx
raffaelefarinaro: Done — added `docs/label-definitions.mdx` to `llmsRootRedirectOnlyDocs` in `docusaurus.config.ts` so the redirect stub is excluded from `llms.txt` / `llms-full.txt` and the canonical Web label-definitions page wins.

PR #352, opened 2026-05-06
Make SparkScan customization partial reusable across platforms

eatsjobs:
Convert the static markdown partial into a parameterized React component that accepts platform-specific API names and features. Pass Web SDK parameters to customize the output for JavaScript differences (property names, delegate API, code language) and hide platform-specific features like hardware trigger and MatrixScan Count that aren't available in Web.

Update all versions (current, 7.6.13, 8.3.1) consistently.


ts1mafei:
Takes some getting used to read React component format instead of markdown. But worth it to reuse description content between platforms 👍

ts1mafei (approved):

PR #367, opened 2026-06-11
Add mobile framework and version selector to doc pages

raffaelefarinaro:
## Why

Feedback from Jenny: on mobile it's difficult or impossible to switch frameworks without going back to the homepage. We deep-link to framework-level pages from the corp site and ~50% of that traffic is mobile.

The only existing path was the easy-to-miss **← Back to main menu** link in the hamburger, behind an entry labeled **SDKs** (nothing says "framework"), and its links dropped users on the Installation page instead of keeping the page they were reading.

## What

**Phones (<768px)** — an in-page framework + version selector, styled like the homepage one, at the top of every doc page (above "On this page"):
- Switching framework keeps the current page and version (e.g. `/sdks/android/id-capture/intro` → `/sdks/ios/id-capture/intro`); pages that don't exist in the target framework fall back via the existing NotFound redirect, same as desktop today.
- Version pill mirrors the navbar version dropdown: Stable/Legacy tags, keeps the current page, offers only 7.6.14/6.28.10 on Xamarin pages.
- Breadcrumbs and the "Back to main menu" button are hidden on mobile (the sidebar already highlights the open page), and the content top padding is reduced 48px → 16px.

**Tablets (768–996px)** — the existing navbar version/framework dropdowns are shown next to the search (they were CSS-hidden below 997px); the in-page selector is hidden in this range. Dropdowns open on tap.

**Desktop (≥997px)** — unchanged.

Also:
- Extracted the desktop dropdown's path-preserving link logic into a shared `useFrameworkItems` hook (the `DropdownNavbarItem` override shrinks by ~190 lines, identical behavior). The mobile SDKs dropdown entries now also keep the current page.
- Fixed the homepage framework selector dropdown being painted over by the SkillsCallout (equal z-index, later DOM order — panel raised to z-index 20).
- `src/theme/DocItem/Layout` is a faithful copy of the stock component with one added line (selector above the mobile TOC).

## Notes for reviewers

- ⚠️ Hiding "Back to main menu" means **Log In / Sign Up are no longer reachable on mobile doc pages** (they only lived in the hamburger primary menu). One-line CSS revert in `navbar-sidebar.scss` if we want them back.
- Framework links are computed after hydration (same as before this PR) so the broken-link checker stays green for pages that don't exist in every framework.

## Testing

- `yarn typecheck` and `yarn build` pass.
- Playwright-verified on iPhone 13 viewport, 768px and 915px tablet widths, and 1440px desktop: switching frameworks/versions navigates correctly and preserves the page, Xamarin filtering works on legacy versions, no console errors.

### DIFF
--- a/src/components/FrameworkSelectorMobile/index.tsx
+++ b/src/components/FrameworkSelectorMobile/index.tsx
@@ -1,0 +1,201 @@
+import React, { useEffect, useRef, useState } from "react";
+import Link from "@docusaurus/Link";
+import { useLocation } from "@docusaurus/router";
+import {
+  useVersions,
+  useActiveDocContext,
+} from "@docusaurus/plugin-content-docs/client";
+import { useDocsPreferredVersion } from "@docusaurus/theme-common";
+import { useFrameworkItems } from "@site/src/utils/useFrameworkItems";
+import { frameworkCards } from "@site/src/components/HomePage/data/frameworkCardsArr";
+import { FrameworksName } from "@site/src/components/constants/frameworksName";
+import { ArrowDown } from "@site/src/components/IconComponents";
+import styles from "./styles.module.css";
+
+// Reuse the homepage framework icons, keyed by display label ("iOS", ".NET Android", ...)
+const iconsByLabel: Record<string, React.ReactNode> = {};
+frameworkCards.forEach((card) => {
+  iconsByLabel[FrameworksName[card.framework as keyof typeof FrameworksName]] =
+    card.icon;
+  card.additional?.forEach((additionalCard) => {
+    iconsByLabel[
+      FrameworksName[additionalCard.framework as keyof typeof FrameworksName]
+    ] = additionalCard.icon;
+  });
+});
+
+// Same categorization as the navbar version dropdown
+const VERSION_CATEGORY: Record<string, string> = {
+  current: "stable",
+};
+const TAG_LABEL: Record<string, string> = {
+  stable: "Stable",
+  beta: "Beta",
+  legacy: "Legacy",
+};
+
+const getVersionMainDoc = (version) =>
+  version.docs.find((doc) => doc.id === version.mainDocId);
+
+function useDropdown(close: () => void) {
+  const fieldRef = useRef<HTMLDivElement>(null);
+  const menuRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    function handleClickOutside(event: MouseEvent | TouchEvent) {
+      if (
+        menuRef.current &&
+        !menuRef.current.contains(event.target as Node) &&
+        fieldRef.current &&
+        !fieldRef.current.contains(event.target as Node)
+      ) {
+        close();
+      }
+    }
+    document.addEventListener("mousedown", handleClickOutside);
+    document.addEventListener("touchstart", handleClickOutside);
+    return () => {
+      document.removeEventListener("mousedown", handleClickOutside);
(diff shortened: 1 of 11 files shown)


src/components/FrameworkSelectorMobile/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+    label: v.label,
+    category: VERSION_CATEGORY[v.name] || "legacy",
+    to: `${
+      (activeDocContext?.alternateDocVersions[v.name] ?? getVersionMainDoc(v))
+        .path
+    }${search}${hash}`,
+    isActive: v === activeVersion,
+  }));
+
+  const versionTag = (category: string) => (
```
ilber, line 114: `getVersionMainDoc` can return undefined here if a version has no doc matching mainDocId, and `.path` would throw. The `??` only guards the first operand — maybe `(... ?? getVersionMainDoc(v))?.path ?? "/"`.

src/components/FrameworkSelectorMobile/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+    <span className={`version-tag version-tag-${category}`}>
+      {TAG_LABEL[category]}
+    </span>
+  );
```
ilber, line 118: `v` comes from useVersions() and activeVersion from useActiveDocContext() — different hooks, so reference equality may never hold and the active version highlight won't show. Compare `v.name === activeVersion?.name` instead.

src/theme/NavbarItem/DropdownNavbarItem/index.js
```diff
@@ -365,16 +195,19 @@ function DropdownNavbarItemMobile({
...
-            mobile
-            isDropdownItem
-            onClick={onClick}
-            activeClassName="menu__link--active"
-            {...childItemProps}
-            key={i}
-          />
-        ))}
+        {effectiveItems.map((childItemProps, i) => {
+          const { sidebarId, ...rest } = childItemProps;
```
ilber, line 206: Worth a short comment on why sidebarId is dropped here while type stays "docsVersion" (so the explicit `to` wins). It's the subtlest behavior change in the PR and the desktop path keeps it.

src/components/FrameworkSelectorMobile/index.tsx
raffaelefarinaro: Good catch — fixed in 0e90280. Resolve `versionDoc` once and use `versionDoc?.path ?? "/"` so a version with no matching main doc falls back to `/` instead of throwing.

src/theme/NavbarItem/DropdownNavbarItem/index.js
raffaelefarinaro: Added a comment at both the mobile and desktop mapping sites in 0e90280: a `docsVersion` item with a `sidebarId` makes Docusaurus resolve the link from the sidebar main doc, which would override our page-preserving `to`; stripping it lets the explicit `to` win.

src/components/FrameworkSelectorMobile/index.tsx
raffaelefarinaro: Right, reference equality never held — the active highlight was silently always off. Switched to `v.name === activeVersion?.name` in 0e90280. Verified at runtime: 8.4.0 option is now active on 8.4.0 pages, 7.6.14 on 7.6.14 pages.

ilber (approved):

PR #371, opened 2026-06-18
Update and fix product availability.

moritzhartmeier:


src/data/products.json
```diff
@@ -199,8 +199,25 @@
...
       "Web": { "version": "n/a" },
-      ".NET iOS": { "version": "n/a" },
-      ".NET Android": { "version": "n/a" }
+      ".NET iOS": { "version": "8.6", "apiUrl": "/sdks/net/ios/barcode-generator/" },
+      ".NET Android": { "version": "8.6", "apiUrl": "/sdks/net/android/barcode-generator/" }
+    }
+  },
+  {
+    "key": "barcode-sequence",
+    "name": "Barcode Sequence",
```
raffaelefarinaro, line 208: We don't list barcode-sequence, this is to be removed

raffaelefarinaro (changes requested):
Only one comment, rest is correct!

src/data/products.json
moritzhartmeier: It is very officially in the release notes since 8.1 already, why would we not list it?

moritzhartmeier: Non-existent intro pages have been removed

raffaelefarinaro (approved):

PR #377, opened 2026-06-25
docs(android): add Unit Testing guide [SDC-31985]

redwarp:
Adds an Android **Unit Testing** page (`docs/sdks/android/unit-testing.mdx`), linked in the Android sidebar next to Agent Skills. Companion to the iOS guide (SDC-31976).

## What it covers
- **Two approaches** (use either/both): isolate the SDK behind your own abstraction, or mock its types.
- **Why** the SDK's modes/sessions/results must be mocked on Android: they're `final` and call native code on construction, so real construction throws `UnsatisfiedLinkError` on the host JVM.
- **MockK and Mockito (inline)** examples, in synced tabs, for every mode the ticket covers — Barcode Capture, MatrixScan (`BarcodeBatch`), ID Capture, SparkScan — plus BarcodePick, `FrameData`, and `Feedback`.
- **Instrumented tests** for what can't run on the JVM: real settings, a real `BarcodeCountView`, and real decoding via `BitmapFrameSource`.
- Tips/pitfalls (e.g. avoid `relaxed = true` on native-backed mocks; background-thread callbacks).

## Validation tooling
The snippet validator gained two changes so these examples compile against the real SDK in CI:
- test/mocking libs (MockK, Mockito, kotlin-test, JUnit, androidx.test) added to the snippet test-bed; `kotlinc` pinned to 2.4.0 (matching `kotlin-test`); `-jvm-target 11` for MockK's inline functions.
- top-level `interface` declarations are hoisted to package scope (Kotlin forbids local interfaces).

All snippets compile: `validate-code-snippets.py kotlin` → **161 passed, 0 failed**. Verified locally and under `act` with the CI toolchain.

### DIFF
--- a/docs/sdks/android/unit-testing.mdx
+++ b/docs/sdks/android/unit-testing.mdx
@@ -1,7 +1,7 @@
 ---
 title: Unit Testing
 sidebar_label: Unit Testing
-description: "Unit-test your scanner listener and callback logic on Android without a camera, by mocking the Scandit Data Capture SDK with MockK or Mockito."
+description: "Unit-test your scanner listener and callback logic on Android without a camera — by isolating the Scandit Data Capture SDK behind your own abstraction, or mocking it with MockK or Mockito."
 toc_max_heading_level: 3
 framework: android
 keywords:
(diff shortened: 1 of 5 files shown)


docs/sdks/android/unit-testing.mdx
```diff
@@ -0,0 +1,656 @@
...
+        session: BarcodeCaptureSession,
+        data: FrameData,
+    ) {
+        scannedData = session.newlyRecognizedBarcode?.data
+        latch.countDown()
+    }
+})
+
+// A bitmap containing a known barcode, e.g. loaded from your test assets.
+val bitmap = BitmapFactory.decodeStream(context.assets.open("test-qr.png"))
```
raffaelefarinaro: `context` is undefined here. The snippet validator injects an ambient `context`, so this compiles in CI, but a reader pasting this into a real `androidTest` has nothing in scope. Suggest grabbing it explicitly first, e.g.

```kotlin
val context = InstrumentationRegistry.getInstrumentation().targetContext
```

before the `assets.open(...)` call (and add the `androidx.test.platform.app.InstrumentationRegistry` import to the hidden block).

docs/sdks/android/unit-testing.mdx
```diff
@@ -0,0 +1,656 @@
...
+val frameSource = BitmapFrameSource.of(bitmap)
+dataCaptureContext.setFrameSource(frameSource)
+frameSource.switchToDesiredState(FrameSourceState.ON)
+
+latch.await(10, TimeUnit.SECONDS)
```
raffaelefarinaro: The `latch.await(...)` return value is ignored. If decoding never fires (invalid/missing license, no match in the image), the latch just times out and the test then asserts on `null` rather than failing on the timeout. Suggest modelling a correct test:

```kotlin
assertTrue(latch.await(10, TimeUnit.SECONDS), "decoding timed out")
```

raffaelefarinaro (commented):
Two small things on the "Verifying Real Decoding" snippet, both reader-facing (CI passes because the validator supplies an ambient binding the reader won't have). Otherwise this looks great.

lucatorella (approved):

docs/sdks/android/unit-testing.mdx
redwarp: Good catch — fixed in 540e4ce7.

One deviation from your suggestion: I used the *test* APK's context rather than `targetContext`, since `test-qr.png` is a fixture that ships in `src/androidTest/assets`, not in the app under test:

```kotlin
val testContext = InstrumentationRegistry.getInstrumentation().context
val bitmap = BitmapFactory.decodeStream(testContext.assets.open("test-qr.png"))
```

I also named it `testContext` instead of `context` so the test-vs-target distinction is visible at the call site, and added a comment saying which assets directory it reads from. Happy to switch to `targetContext` if you meant the reader to keep the fixture in the app's own assets.

docs/sdks/android/unit-testing.mdx
redwarp: Agreed, applied verbatim in 540e4ce7:

```kotlin
assertTrue(latch.await(10, TimeUnit.SECONDS), "decoding timed out")
```

Worth noting this also improves the failure message: previously a timeout surfaced as `expected:<Scandit> but was:<null>`, which reads like a decoding bug when the usual cause is a missing or invalid license key.

lucatorella (approved):

PR #381, opened 2026-07-03
Add UIViewRepresentable integration to the SparkScan SwiftUI guide

ts1mafei:
## What

Expands the iOS SparkScan [Get Started With SwiftUI](docs/sdks/ios/sparkscan/get-started-with-swift-ui.md) guide with a new section documenting the pure-SwiftUI integration path: wrapping `SparkScanView` directly with `UIViewRepresentable` and a coordinator, as an alternative to the existing `withSparkScan` view modifier.

Part of [SDC-31151](https://scandit.atlassian.net/browse/SDC-31151).

## Structure

The new section is placed at the bottom of the page, below the existing `UIViewControllerRepresentable`-based content, and follows the page's existing style (step-by-step sections, small code chunks, API links):

- **Create the UIViewRepresentable** — the wrapper struct; lifecycle driven by an `isActive` flag instead of `viewWillAppear`/`viewWillDisappear`; hit-test forwarding for the floating UI
- **Create the Coordinator** — owns the long-lived SparkScan objects; forwards `SparkScanListener` and `SparkScanFeedbackDelegate` to SwiftUI closures (with the threading contracts called out)
- **Update Settings While Scanning** — diff-and-apply on the existing settings instance so other customizations are preserved; notes that any `SparkScanSettings` property can be forwarded the same way
- **Use the Scanner in Your SwiftUI App** — state-driven usage with a symbology toggle example

The intro notes the trade-off: `withSparkScan` is often the easier path (especially for more advanced configurations), while `UIViewRepresentable` can feel more natural in a pure SwiftUI codebase.

## Notes

- The code mirrors a sample implementation that compiles against the current SDK (`ListBuildingSampleSwiftUIRepresentable`, in review for `data-capture-sdk`).
- Edited `docs/` (next version) only; versioned docs untouched.
- `npm run build` passes; the section renders under the `next` version route.

### DIFF
--- a/docs/sdks/ios/sparkscan/get-started-with-swift-ui.md
+++ b/docs/sdks/ios/sparkscan/get-started-with-swift-ui.md
@@ -148,3 +148,222 @@ struct ScanningView: View {
     }
 }
 ```
+
+## Alternative: Integrate Using UIViewRepresentable
+
+If you prefer to avoid view controllers entirely, you can wrap the [`SparkScanView`](https://docs.scandit.com/data-capture-sdk/ios/barcode-capture/api/ui/spark-scan-view.html) directly with [`UIViewRepresentable`](https://developer.apple.com/documentation/swiftui/uiviewrepresentable) and a coordinator.
+
+Using the `withSparkScan` view modifier described above is often the easier integration path, especially when handling more advanced configurations. The bonus of `UIViewRepresentable` is that it can feel more natural in a pure SwiftUI codebase:
+
+- **State flows down**: SwiftUI state (such as the enabled symbologies, or whether scanning is active) is pushed into the running scanner whenever it changes.
+- **Events flow up**: scan results and feedback decisions are forwarded to your SwiftUI code through closures.
+
+### Create the UIViewRepresentable
+
+Create a `SparkScanScanner` struct that creates the [`SparkScanView`](https://docs.scandit.com/data-capture-sdk/ios/barcode-capture/api/ui/spark-scan-view.html) in `makeUIView` and pushes SwiftUI state changes to the scanner in `updateUIView`.
+
+Since there is no view controller in this approach, [`SDCSparkScanView.prepareScanning`](https://docs.scandit.com/data-capture-sdk/ios/barcode-capture/api/ui/spark-scan-view.html#method-scandit.datacapture.barcode.spark.ui.SparkScanView.PrepareScanning) and [`SDCSparkScanView.stopScanning`](https://docs.scandit.com/data-capture-sdk/ios/barcode-capture/api/ui/spark-scan-view.html#method-scandit.datacapture.barcode.spark.ui.SparkScanView.StopScanning) are driven by an `isActive` flag instead of the `viewWillAppear` and `viewWillDisappear` callbacks:
+
+```swift
+import ScanditBarcodeCapture
+import SwiftUI
+
+struct SparkScanScanner: UIViewRepresentable {
+    // State flows down from SwiftUI.
+    let dataCaptureContext: DataCaptureContext
+    let symbologies: Set<Symbology>
+    let isActive: Bool
+    // Events flow up to your SwiftUI code.
+    let onScan: (Barcode) -> Void
+    let feedbackFor: (Barcode) -> SparkScanBarcodeFeedback?
+
+    func makeCoordinator() -> SparkScanScannerCoordinator {
+        let settings = SparkScanSettings()
+        symbologies.forEach { settings.set(symbology: $0, enabled: true) }
+
+        return SparkScanScannerCoordinator(
+            dataCaptureContext: dataCaptureContext,
+            settings: settings,
+            onScan: onScan,
+            feedbackFor: feedbackFor
+        )
+    }
+
+    func makeUIView(context: Context) -> UIView {
+        // SparkScanView attaches itself to the parent view passed at creation.
+        let parentView = ParentView()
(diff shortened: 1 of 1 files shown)


docs/sdks/ios/sparkscan/get-started-with-swift-ui.md
```diff
@@ -147,4 +147,223 @@ struct ScanningView: View {
...
+
+        return parentView
+    }
+
+    func updateUIView(_ uiView: UIView, context: Context) {
+        context.coordinator.update(onScan: onScan, feedbackFor: feedbackFor)
+        // Push the current SwiftUI state into the running scanner.
+        context.coordinator.applySymbologiesIfNeeded(symbologies)
+        if isActive {
+            context.coordinator.sparkScanView?.prepareScanning()
```
marcozoffoli-scandit: `updateUIView` runs on every SwiftUI state change — in the usage example below, that includes every scanned barcode being appended to `scannedBarcodes`. As written, `prepareScanning()` is therefore re-invoked once per scan while the scanner is active.

If a redundant `prepareScanning()` is a guaranteed no-op in the SDK this is harmless, but it might be safer not to rely on readers knowing that. One option is to have the coordinator remember the last applied value and only call `prepareScanning()`/`stopScanning()` on transitions, e.g.:

```swift
if isActive != context.coordinator.isScanningActive {
    context.coordinator.isScanningActive = isActive
    isActive ? context.coordinator.sparkScanView?.prepareScanning()
             : context.coordinator.sparkScanView?.stopScanning()
}
```

That would also make this consistent with the diff-before-apply pattern used in `applySymbologiesIfNeeded` below.

docs/sdks/ios/sparkscan/get-started-with-swift-ui.md
```diff
@@ -147,4 +147,223 @@ struct ScanningView: View {
...
+        } else {
+            context.coordinator.sparkScanView?.stopScanning()
+        }
+    }
+}
```
marcozoffoli-scandit, line 230: Small suggestion: when the scanner view is removed from the hierarchy (navigation pop, conditional view, etc.) while `isActive` is still `true`, `stopScanning()` is never called and cleanup relies on deallocation. Adding a `dismantleUIView` would model the full lifecycle for readers:

```swift
static func dismantleUIView(_ uiView: UIView, coordinator: SparkScanScannerCoordinator) {
    coordinator.sparkScanView?.stopScanning()
}
```

docs/sdks/ios/sparkscan/get-started-with-swift-ui.md
```diff
@@ -147,4 +147,223 @@ struct ScanningView: View {
...
+        DispatchQueue.main.async {
+            self.onScan(barcode)
+        }
+    }
+}
+
+extension SparkScanScannerCoordinator: SparkScanFeedbackDelegate {
+    // Invoked on a background queue: the closure must only access
+    // thread-safe state.
+    func feedback(for barcode: Barcode) -> SparkScanBarcodeFeedback? {
```
marcozoffoli-scandit, line 308: Minor thread-safety note: this delegate method reads `self.feedbackFor` from a background queue, while `update(onScan:feedbackFor:)` reassigns the same property from the main thread on every render — so there's an unsynchronized read/write on the coordinator's stored closures themselves (slightly at odds with the comment above, which warns about state accessed *inside* the closure).

For a docs example this may well be an acceptable simplification, but it could be worth either protecting the closure properties with a small lock, or extending the comment to acknowledge the simplification so readers don't copy it into more complex setups unaware.

marcozoffoli-scandit (commented):
Thanks for adding this — nice addition to the SwiftUI guide, and the threading callouts are a great touch. I left a few small, non-blocking suggestions inline, mostly around making the example as copy-paste-robust as possible for readers.

docs/sdks/ios/sparkscan/get-started-with-swift-ui.md
ts1mafei: 🤖 Good catch — and your hedge was actually too generous: a redundant `prepareScanning()` is *not* a no-op. In the SDK it unconditionally resets the view state to Idle and re-applies camera settings (`SDCSparkScanViewManager.mm`, `prepareScanning`), so calling it per render can interrupt active scanning after every recognized barcode. Adopted your transition guard (via `coordinator.isScanningActive`) in fa032196 — it also keeps the example consistent with the diff-before-apply pattern in `applySymbologiesIfNeeded`. The change is verified in the compiling companion sample (`ListBuildingSampleSwiftUIRepresentable` in data-capture-sdk).

docs/sdks/ios/sparkscan/get-started-with-swift-ui.md
ts1mafei: 🤖 Agreed — added `dismantleUIView` calling `stopScanning()` in fa032196, so the teardown path is modeled too.

docs/sdks/ios/sparkscan/get-started-with-swift-ui.md
ts1mafei: 🤖 Fair point — the read/write on the stored closures themselves is indeed unsynchronized. Went with your lighter option to keep the get-started example lean: the comment now explicitly acknowledges the simplification and tells readers to add locking around the stored closures if their app replaces them frequently (fa032196).

marcozoffoli-scandit (approved):

PR #384, opened 2026-07-09
Add API Reference fallback link and fix docs search event capture

eugenia-scandit:
Most zero-result searches are exact API symbol names because the data-capture-sdk tree is excluded from the crawler index. Until it is re-included, the no-results screen now links to the API reference's built-in search with the user's query preserved, framework-aware


src/theme/SearchBar/index.js
```diff
@@ -244,14 +262,29 @@ function DocSearch({ contextualSearch, externalUrlRegex, ...props }) {
             transformItems={transformItems}
             hitComponent={Hit}
             transformSearchClient={transformSearchClient}
-            onStateChange={handleStateChange}
+            getMissingResultsUrl={({ query }) => {
+              // Most zero-result queries are exact API symbol names that the
+              // main index doesn't contain (data-capture-sdk tree excluded) -
+              // offer the API reference's built-in search as a fallback.
+              const fw = (currentFramework || '/sdks/web')
```
moritzhartmeier: This produces broken links for .NET. Seems like two issues:

1. currentFramework's regex (/\/sdks\/([\w-]+(?:\/android)?)/, on line 72) only special-cases /android — /ios is never captured, so the net/ -> net. replacement doesn't happen thereon .NET iOS pages currentFramework collapses to bare net, which also 404s.
2. the replacement should be net/ -> dotnet. because now even net/android (which the regex correctly captures) maps to net.android, and the live convention for the api docs is dotnet.android/dotnet.ios

Claude suggests replacing the .replace() chain with an explicit framework→segment map (e.g. {'net/android': 'dotnet.android', 'net/ios': 'dotnet.ios', web: 'web', ...}) rather than string surgery — it's the only way to express the net→dotnet rename and fixes the iOS gap in one place.

moritzhartmeier (changes requested):

src/theme/SearchBar/index.js
```diff
@@ -244,14 +332,27 @@ function DocSearch({ contextualSearch, externalUrlRegex, ...props }) {
...
-            {...(props.searchPagePath && {
-              resultsFooterComponent,
-            })}
+            getMissingResultsUrl={({ query }) => {
+              // Most zero-result queries are exact API symbol names that the
+              // main index doesn't contain (data-capture-sdk tree excluded) -
+              // offer the API reference's built-in search as a fallback.
+              const fw = (currentFramework || '/sdks/web')
+                .replace('/sdks/', '')
+                .replace('net/', 'net.');
```
moritzhartmeier: This still maps to net. instead of dotnet.  the live API reference convention is dotnet.ios/dotnet.android (see eg. https://docs.scandit.com/data-capture-sdk/dotnet.android/core/api.html)

src/theme/SearchBar/index.js
```diff
@@ -155,9 +207,20 @@ function DocSearch({ contextualSearch, externalUrlRegex, ...props }) {
...
     (items) => {
-      const filteredItems = items.filter((elem) =>
-        elem.url.includes(currentFramework)
-      );
+      // API pages live under /data-capture-sdk/<fw>/, not /sdks/<fw>/, so match
+      // them by their own framework segment (net.ios -> net/ios). On a
+      // framework-less page (home, /hosted/, concept pages) fall back to the
+      // most-used framework so API results aren't duplicated across all SDKs.
+      const fwToken = currentFramework.replace(/^\/sdks\//, "");
+      const apiFwTarget = (fwToken || API_FALLBACK_FRAMEWORK).toLowerCase();
```
moritzhartmeier, line 234: Same net/dotnet mismatch as the fallback link below: apiFwTarget here will be net/ios for .NET, but API page URLs use dotnet.ios.

moritzhartmeier (changes requested):

moritzhartmeier (approved):

PR #385, opened 2026-07-09
Fix Agent Skills callout framework linking and add ID scanning skills

raffaelefarinaro:
## Summary

Addresses three issues with the Agent Skills callout/badge and fills a gap in the tracked skills catalog.

### 1. "More info" now keeps the framework
The callout's **More info →** link used to jump to iOS instead of staying on the framework the reader was on.

- **Shared callout (in-page):** now derives the framework from the current page path.
- **Homepage banner:** the homepage switches frameworks via `history.pushState`, which React Router never observes — so `useLocation()` stayed stale and the link kept its initial (iOS) value even for `?framework=web`. The banner now resolves the target **at click time** from the live URL / `localStorage`. Also added an alias map for the homepage's own identifiers (`react` → `react-native`, `netIos` → `net-ios`, `netAndroid` → `net-android`), which otherwise fell back to iOS.

### 2. Hide the badge on Titanium & Linux
No skills exist for those platforms, so the callout no longer renders under `/sdks/titanium/` or `/sdks/linux/`.

### 3. Track the ID scanning skills (and reconcile the whole catalog)
`skills.json` was significantly out of date vs. [`scandit/skills`](https://github.com/scandit/skills). Reconciled it so **every** published skill is tracked (verified bidirectionally — no missing, no dangling slugs):

- **ID Capture** across all 9 frameworks + the `.NET MAUI` variant.
- **ID Bolt** listed on the Web Agent Skills page, plus a **dedicated skill callout on the hosted ID Bolt docs** (`/hosted/id-bolt/*`), which live outside the `/sdks/` product routes and previously showed only the generic product-picker callout.
- Previously-missing entries now added: `barcode-capture-ios`, `matrixscan-pick`, the missing `.NET iOS/Android` skills for sparkscan/matrixscan-*/label-capture, and the full MAUI block.

## Files changed
- `src/data/skills.json` — full catalog reconciliation
- `src/components/SkillsCallout/index.tsx` — framework-keeping "More info" + new `skill` variant
- `src/components/SkillsPage/index.tsx` — id-capture/matrixscan-pick descriptions + `url`/`description` field support (for ID Bolt)
- `src/theme/DocItem/Content/index.tsx` — hide on titanium/linux + ID Bolt hosted-docs callout

## Testing
- `docusaurus build` passes.
- Verified in built HTML: ID Bolt callout renders on `/hosted/id-bolt/getting-started` (and is correctly absent on `release-notes`).
- Unit-tested the framework→URL resolution across all homepage identifiers (`web`, `react`, `netIos`, `netAndroid`, unknowns → iOS fallback).
- Confirmed the homepage `?framework=web` "More info" fix live in the dev server.

### DIFF
--- a/src/components/SkillsCallout/index.tsx
+++ b/src/components/SkillsCallout/index.tsx
@@ -3,7 +3,7 @@ import { useLocation } from '@docusaurus/router';
 
 import skillsData from '@site/src/data/skills.json';
 import productsData from '@site/src/data/products.json';
-import { parseSdksRoute, frameworkToSlug } from '../utils/frameworks';
+import { parseSdksRoute, frameworkToSlug, FRAMEWORK_STORAGE_KEY } from '../utils/frameworks';
 import { capturePostHogEvent } from './analytics';
 import InstallTabs from './InstallTabs';
 import styles from './styles.module.css';
@@ -111,7 +111,7 @@ function resolveLiveBannerMoreInfoUrl(): string {
   const fromQuery = new URLSearchParams(window.location.search).get('framework');
   const fromStorage = (() => {
     try {
-      return window.localStorage.getItem('framework');
+      return window.localStorage.getItem(FRAMEWORK_STORAGE_KEY);
     } catch {
       return null;
     }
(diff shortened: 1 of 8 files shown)


src/components/SkillsCallout/index.tsx
```diff
@@ -61,17 +68,58 @@ const QUERY_FRAMEWORK_TO_PATH: Record<string, string> = {
...
+// The homepage swaps frameworks with history.pushState, which does NOT update
+// React Router — so useLocation() (and any href derived from it) goes stale.
+// For the banner we therefore resolve the selected framework from the live URL
+// and localStorage at click time instead of trusting the rendered value.
+function resolveLiveBannerMoreInfoUrl(): string {
+  if (typeof window === 'undefined') return '/sdks/ios/agent-skills';
+  const fromQuery = new URLSearchParams(window.location.search).get('framework');
+  const fromStorage = (() => {
+    try {
+      return window.localStorage.getItem('framework');
```
ilber: 🤖 Correctness here hinges on the homepage selector writing exactly `localStorage["framework"]`. That contract is not visible in this diff — if the homepage ever renames the key, the banner silently reverts to the iOS fallback with no error. Worth pinning it with a shared constant or a comment naming the writer.

src/components/SkillsCallout/index.tsx
```diff
@@ -138,12 +188,67 @@ const SharedBody: React.FC<SharedBodyProps> = ({
...
-    const sharedFrameworkSlug = getSharedFrameworkSlug(search);
-    const sharedMoreInfoUrl = getSharedMoreInfoUrl(search);
+    const sharedFrameworkSlug = getSharedFrameworkSlug(pathname, search);
+    const sharedMoreInfoUrl = getSharedMoreInfoUrl(pathname, search);
+    // On the homepage banner, the framework is changed via history.pushState
+    // (no router update), so recompute the target from the live URL on click.
+    const onMoreInfoClick: React.MouseEventHandler<HTMLAnchorElement> | undefined = banner
+      ? (e) => {
+          if (
+            e.button !== 0 ||
```
ilber: 🤖 Modified/middle clicks still leak the stale iOS URL. This guard lets the browser fall through to `href` on cmd/ctrl/shift/middle-click, but on the homepage `href` = `sharedMoreInfoUrl` derived from stale `useLocation()` — the exact bug the click-time resolver fixes for plain left-click. So cmd-click "More info" on `?framework=web` opens iOS agent-skills in a new tab. Non-blocking, but consider resolving `href` live too.

src/components/SkillsCallout/index.tsx
raffaelefarinaro: Good catch — fixed in ad5571d5. Added a shared `FRAMEWORK_STORAGE_KEY` constant in `src/components/utils/frameworks.ts` and routed both the homepage writers (`Frameworks.tsx`, `CardAdditional.tsx`, `FrameworkCard.tsx`) and this reader through it, so the contract is pinned in one place.

src/components/SkillsCallout/index.tsx
raffaelefarinaro: Fixed in ad5571d5. The banner now refreshes the anchor `href` from the live URL on `mousedown` and `focus` (both fire before the browser navigates), so cmd/ctrl/shift/middle-click and keyboard activation open the correct framework instead of the stale iOS href.

ilber (approved):

PR #392, opened 2026-07-16
Route search by framework and version named in the query

eugenia-scandit:
Search results are filtered to the framework/version of the page the user is on. When the user names one explicitly in the query, that should win - so "barcode capture ios" from a web page returns iOS results.

- Framework: a framework typed in the query (ios, android, flutter, web, react native, capacitor, cordova, titanium, .net ios/android) overrides the page framework; two named frameworks return both. Nothing typed keeps the page framework. On a framework-less page (home, /hosted/, concept pages) guides fall back to the most-used framework (web) rather than every framework's guide, while framework-agnostic pages (/hosted/, /id-documents/) stay findable.

- Version: a version named in the query (version 6, v6, 6.x, 6.28, sdk 6) rewrites the contextual docusaurus_tag filter to the newest tag on that major line - 6 to docs-default-6.28.11, 7 to docs-default-7.6.14, 8 to current.

The typed framework word is left in the query; removeWordsIfNoResults drops it when it does not match page text, so ranking is unaffected.

### DIFF
--- a/src/theme/SearchBar/index.js
+++ b/src/theme/SearchBar/index.js
@@ -83,19 +83,16 @@ function frameworksInQuery(query) {
   return found;
 }
 // Major version typed in the query -> the newest docusaurus_tag on that line.
-// Update when a major version's newest patch changes or a new line ships.
-const VERSION_TAG_BY_MAJOR = {
-  "6": "docs-default-6.28.11",
-  "7": "docs-default-7.6.14",
-  "8": "docs-default-current",
-};
-function versionTagInQuery(query) {
+// The map is derived from docsVersions in docusaurus.config.ts and passed in via
+// customFields (versionTagByMajor), so it can never drift from the real versions.
+const EMPTY_VERSION_MAP = {};
+function versionTagInQuery(query, versionTagByMajor) {
   const q = (query || "").toLowerCase();
   // "version 6" / "ver 6" / "v6" / "sdk 6", or a dotted form like "6.28" / "6.x".
   const m =
     q.match(/\b(?:version|ver|v|sdk)\s*\.?\s*(\d+)\b/) ||
     q.match(/\b(\d+)\.(?:x|\d+)\b/);
-  return m ? VERSION_TAG_BY_MAJOR[m[1]] || null : null;
+  return m ? versionTagByMajor[m[1]] || null : null;
 }
 function rewriteVersionTag(facetFilters, targetTag) {
   const swap = (f) => {
@@ -165,7 +162,10 @@ function mergeFacetFilters(f1, f2) {
   return [...normalize(f1), ...normalize(f2)];
 }
 function DocSearch({ contextualSearch, externalUrlRegex, ...props }) {
-  const { siteMetadata } = useDocusaurusContext();
+  const { siteMetadata, siteConfig } = useDocusaurusContext();
+  // Version-routing map derived at build time from docsVersions (see config).
+  const versionTagByMajor =
+    siteConfig.customFields?.versionTagByMajor || EMPTY_VERSION_MAP;
   const processSearchResultUrl = useSearchResultUrlProcessor();
   const contextualSearchFacetFilters = useAlgoliaContextualFacetFilters();
   const configFacetFilters = props.searchParameters?.facetFilters ?? [];
@@ -385,7 +385,7 @@ function DocSearch({ contextualSearch, externalUrlRegex, ...props }) {
         latestQueryRef.current = query || "";
         // A version typed in the query overrides the page's version: swap the
         // contextual docusaurus_tag filter to the newest tag for that major.
-        const targetTag = versionTagInQuery(query);
+        const targetTag = versionTagInQuery(query, versionTagByMajor);
         const effectiveRequests =
           targetTag && Array.isArray(requests)
             ? requests.map((r) => {
@@ -414,7 +414,7 @@ function DocSearch({ contextualSearch, externalUrlRegex, ...props }) {
       };
       return searchClient;
     },
-    [siteMetadata.docusaurusVersion, captureSearchDebounced]
+    [siteMetadata.docusaurusVersion, captureSearchDebounced, versionTagByMajor]
   );
   useDocSearchKeyboardEvents({
     isOpen,
--- a/src/theme/SearchBar/index.js
+++ b/src/theme/SearchBar/index.js
@@ -69,7 +69,10 @@ const QUERY_FRAMEWORK_TOKENS = [
   { re: /\bcapacitor\b/, fw: "capacitor" },
   { re: /\bcordova\b/, fw: "cordova" },
   { re: /\btitanium\b/, fw: "titanium" },
-  { re: /\bweb\b/, fw: "web" },
+  // "web" needs stronger context ("web sdk", "on/for/in/using web"). Bare "web"
+  // is a common English word ("web view", "web socket") and would hijack
+  // routing on ordinary queries, so an accompanying cue is required.
+  { re: /\bweb\s+sdk\b|\b(?:on|for|in|using)\s+web\b/, fw: "web" },
 ];
 function frameworksInQuery(query) {
   let q = ` ${(query || "").toLowerCase()} `;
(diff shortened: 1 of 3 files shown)


src/theme/SearchBar/index.js
```diff
@@ -55,6 +55,59 @@ function sdkFrameworkToApi(sdkToken) {
...
+    if (re.test(q)) {
+      if (!found.includes(fw)) found.push(fw);
+      q = q.replace(re, " "); // consume so a longer token isn't re-matched
+    }
+  }
+  return found;
+}
+// Major version typed in the query -> the newest docusaurus_tag on that line.
+// Update when a major version's newest patch changes or a new line ships.
+const VERSION_TAG_BY_MAJOR = {
```
moritzhartmeier: This map hardcodes patch versions (6.28.11, 7.6.14) as well as mappings to current ("8") that mirror versions.json today, but nothing keeps them in sync.

I see two options:
1) Derive these strings from versions.json and the current versions as defined in docusaurus.config.ts
2) Update scripts/update-version.py to ensure these versions get bumped when needed

moritzhartmeier (dismissed):

src/theme/SearchBar/index.js
eugenia-scandit: Went with option 1, deriving it

src/theme/SearchBar/index.js
```diff
@@ -55,6 +55,56 @@ function sdkFrameworkToApi(sdkToken) {
...
+function rewriteVersionTag(facetFilters, targetTag) {
+  const swap = (f) => {
+    if (typeof f === "string") {
+      return f.indexOf("docusaurus_tag:") === 0
+        ? `docusaurus_tag:${targetTag}`
+        : f;
+    }
+    return Array.isArray(f) ? f.map(swap) : f;
+  };
+  return swap(facetFilters);
```
lucatorella, line 113: **[correctness, CONFIRMED]** `rewriteVersionTag` overwrites *every* `docusaurus_tag` entry, including `docusaurus_tag:default`.

Docusaurus contextual search builds `facetFilters` with multiple `docusaurus_tag` entries (an OR array plus `docusaurus_tag:default` for non-versioned pages). This function replaces every string starting with `docusaurus_tag:` with the single target tag, collapsing the OR and clobbering the `default` tag. When a user types a version, framework-agnostic/non-versioned pages that were only tagged `default` no longer match the facet and drop out of results entirely (e.g. "id capture v6" can return zero of the non-versioned id-documents pages that would otherwise match). Also affects line 100.

src/theme/SearchBar/index.js
```diff
@@ -55,6 +55,56 @@ function sdkFrameworkToApi(sdkToken) {
...
+}
+// Major version typed in the query -> the newest docusaurus_tag on that line.
+// The map is derived from docsVersions in docusaurus.config.ts and passed in via
+// customFields (versionTagByMajor), so it can never drift from the real versions.
+const EMPTY_VERSION_MAP = {};
+function versionTagInQuery(query, versionTagByMajor) {
+  const q = (query || "").toLowerCase();
+  // "version 6" / "ver 6" / "v6" / "sdk 6", or a dotted form like "6.28" / "6.x".
+  const m =
+    q.match(/\b(?:version|ver|v|sdk)\s*\.?\s*(\d+)\b/) ||
```
lucatorella: **[correctness, CONFIRMED]** Version detection is over-eager and silently rewrites the facet.

`versionTagInQuery`'s patterns — `(?:version|ver|v|sdk)\s*\.?\s*\d+` and any dotted `N.M` — match incidental tokens in ordinary queries (a spec/dimension/error code, e.g. "6.x", "7.1", "sdk 6"). This silently rewrites the contextual `docusaurus_tag` facet to an older docs version (e.g. `docs-default-6.28.11`) with no indication to the user that the version was switched. Also affects line 94.

src/theme/SearchBar/index.js
```diff
@@ -232,18 +288,42 @@ function DocSearch({ contextualSearch, externalUrlRegex, ...props }) {
...
+        }
+        if (guideSegments) {
+          return guideSegments.some((seg) => url.includes(seg));
+        }
+        if (currentFramework) {
+          return url.includes(currentFramework);
+        }
+        // Framework-less page: the most-used framework's guides only (not every
+        // framework's), plus pages tied to no framework so they stay findable.
+        if (url.includes("/sdks/")) {
```
lucatorella: **[correctness, CONFIRMED]** Framework-less-page searches now return web-only guides instead of guides from all SDKs.

On main, a search from any framework-less page (homepage, `/hosted/`, concept pages) returned guides from every SDK because `url.includes(currentFramework)` with `currentFramework=""` matched everything. This branch instead returns only `/sdks/web/` guides for any `/sdks/` URL — a user primarily working with Android/iOS/Flutter who searches a non-framework term from the homepage now silently loses the other SDKs' guides. This also runs counter to AGENTS.md's "all SDK platforms are treated equally / no platform priority".

src/theme/SearchBar/index.js
```diff
@@ -232,18 +288,42 @@ function DocSearch({ contextualSearch, externalUrlRegex, ...props }) {
...
+        ? queriedFrameworks.map((fw) => `/sdks/${fw}/`)
+        : null;
       const filteredItems = items.filter((elem) => {
         const url = elem.url || "";
         const apiMatch = url.match(/\/data-capture-sdk\/([^/]+)\//);
         if (apiMatch) {
-          return apiFrameworkToSdk(apiMatch[1]) === apiFwTarget;
+          return apiFwTargets.includes(apiFrameworkToSdk(apiMatch[1]));
+        }
+        if (guideSegments) {
```
lucatorella: **[correctness, CONFIRMED]** Framework-agnostic pages are filtered out once a framework token is detected in the query, contradicting the code's own intent to keep them findable.

Example: searching "ios id document" sets `hasQueriedFramework=true` and `guideSegments=["/sdks/ios/"]`; a concept page like `/id-documents/` or `/hosted/` matches no guide segment and is dropped by `guideSegments.some(...)` — adding a platform word to narrow a search makes framework-agnostic pages disappear entirely (they only survive the framework-less branch at lines 323-326). Also affects line 316.

src/theme/SearchBar/index.js
```diff
@@ -55,6 +55,56 @@ function sdkFrameworkToApi(sdkToken) {
...
+  { re: /\breact[\s-]?native\b/, fw: "react-native" },
+  { re: /\b(?:dot)?net[\s./]*ios\b/, fw: "net/ios" },
+  { re: /\b(?:dot)?net[\s./]*android\b/, fw: "net/android" },
+  { re: /\bios\b/, fw: "ios" },
+  { re: /\bandroid\b/, fw: "android" },
+  { re: /\bflutter\b/, fw: "flutter" },
+  { re: /\bcapacitor\b/, fw: "capacitor" },
+  { re: /\bcordova\b/, fw: "cordova" },
+  { re: /\btitanium\b/, fw: "titanium" },
+  { re: /\bweb\b/, fw: "web" },
```
lucatorella: **[correctness, CONFIRMED]** The `\bweb\b` query token hijacks framework routing whenever the common English word "web" appears.

Unlike "ios"/"android", "web" is a common substring word in normal queries ("web view", "web socket", "web app"). An Android developer on an Android guide page searching "web view" trips `\bweb\b`, `hasQueriedFramework` becomes true, and results are forcibly narrowed to `/sdks/web/` guides and the web API reference — the Android results they wanted disappear.

lucatorella (commented):
Automated code review (Claude Code, high effort). 6 findings below — 5 inline, plus one that anchors to unchanged code outside this diff's hunks so it's noted here instead:

**[correctness, CONFIRMED] `ResultsFooter`'s fallback note can contradict the actual results shown** (`src/theme/SearchBar/index.js`, `ResultsFooter` / `showApiFallbackNote`, around line 85 on `main`).

On a framework-less page, `currentFramework` is `""`. A user searches "barcode capture ios"; `transformItems` filters API hits to iOS (via `apiFwTargets = queriedFrameworks`), but `showApiFallbackNote = !currentFramework && hasApiResults` is still true, so the footer renders "Showing the web API reference - open a specific SDK's docs for another platform" even though the results shown are iOS, not web.

Overall theme of the inline findings: the new version/framework query-inference logic false-positives on ordinary query text and silently narrows or rewrites the search facets, in several cases contradicting comments in the same file promising broader/framework-agnostic results.

src/theme/SearchBar/index.js
lucatorella: this sounds like a potential problem

src/theme/SearchBar/index.js
lucatorella: ignore

src/theme/SearchBar/index.js
lucatorella: ignore

src/theme/SearchBar/index.js
eugenia-scandit: This is intended behavior to avoid spamming user with results for all the frameworks. All other points are fair catches, I will push fixes

lucatorella (approved):
Approving after review — inline findings addressed in commit aca6b32 ("Harden query-based framework/version routing from review feedback").

PR #403, opened 2026-07-24
[SDC-32633] Document Web SDK UMD builds

anthonypenna:
Adds an "Install via UMD build" section to the Web SDK installation page, covering how to load the SDK from plain `<script>` tags via browser globals, without a bundler — complementing the existing npm and ESM/CDN options.

Tested locally with `yarn build`.

Note for reviewers: the UMD build ships in **8.6.0** (SDC-31903, fix version `8.6.0-beta.1`). It is not in 8.5.x — verified not cherry-picked to `release/8.5` — so the `/umd` CDN paths in the example won't resolve until 8.6.0 is released. Please don't publish this page before the 8.6.0 release.

Closes SDC-32633

### DIFF
--- a/docs/sdks/web/add-sdk.md
+++ b/docs/sdks/web/add-sdk.md
@@ -267,6 +267,120 @@ In alternative to jsdeliver, unpkg can be used:
 </html>
 ```
 
+## Install via UMD build (plain `<script>` tags)
+
+Alongside the ES module (ESM) build, every Scandit package ships a [UMD](https://github.com/umdjs/umd) build. This lets you load the SDK with plain `<script>` tags, without a bundler or an [import map](#install-via-cdn). Each package exposes its exports on a browser global:
+
+| Package | Browser global |
+| --- | --- |
+| `@scandit/web-datacapture-core` | `SDCCore` |
+| `@scandit/web-datacapture-barcode` | `SDCBarcode` |
+| `@scandit/web-datacapture-id` | `SDCId` |
+| `@scandit/web-datacapture-label` | `SDCLabel` |
+| `@scandit/web-datacapture-parser` | `SDCParser` |
+
+The UMD bundle for each package lives at `@scandit/web-datacapture-<module>/umd/index.js`. On a CDN you can also point to the bare `umd` subpath (for example `.../web-datacapture-core/umd`) and it resolves to `index.js` via the directory index.
+
+:::caution Load order matters
+Load `@scandit/web-datacapture-core` **first**. The feature packages (barcode, id, label, parser) resolve the `SDCCore` global when they load, so their `<script>` tags must come after core's.
+:::
+
+:::note
+A classic `<script>` (unlike `<script type="module">`) does not support top-level `await`, so wrap the setup code in an async function as shown below.
+:::
+
+### Complete UMD Example
+
+```html
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title>Scandit UMD Simple sample</title>
+    <!-- Load core first, then the feature packages that depend on it. -->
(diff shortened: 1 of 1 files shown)


docs/sdks/web/add-sdk.md
```diff
@@ -267,6 +267,120 @@ In alternative to jsdeliver, unpkg can be used:
...
+
+:::caution Load order matters
+Load `@scandit/web-datacapture-core` **first**. The feature packages (barcode, id, label, parser) resolve the `SDCCore` global when they load, so their `<script>` tags must come after core's.
+:::
+
+:::note
+A classic `<script>` (unlike `<script type="module">`) does not support top-level `await`, so wrap the setup code in an async function as shown below.
+:::
+
+### Complete UMD Example
```
sebastien-roch: Can't we just keep the interesting part in this section? a complete example seems superfluous and gives a lot of space to UMD in this page.

docs/sdks/web/add-sdk.md
```diff
@@ -267,6 +267,120 @@ In alternative to jsdeliver, unpkg can be used:
 </html>
 ```
 
+## Install via UMD build (plain `<script>` tags)
+
+Alongside the ES module (ESM) build, every Scandit package ships a [UMD](https://github.com/umdjs/umd) build. This lets you load the SDK with plain `<script>` tags, without a bundler or an [import map](#install-via-cdn). Each package exposes its exports on a browser global:
```
sebastien-roch: We should mention that using UMD leads to longer startup time because of larger files to download (it cannot be optimised like a bundle consuming our ES modules), so it should only be used when ES module is not possible.

eatsjobs: I would also suggest the use of https://github.com/guybedford/es-module-shims#usage before using UMD bundle at all. And considering the browser support of ESM https://caniuse.com/?search=ESM

UMD should really be used as last resort @anthonypenna also

docs/sdks/web/add-sdk.md
```diff
@@ -267,6 +267,120 @@ In alternative to jsdeliver, unpkg can be used:
...
+        Symbology,
+        SymbologyDescription,
+      } = SDCBarcode;
+
+      (async () => {
+        let view = new DataCaptureView();
+        view.connectToElement(document.getElementById("app"));
+        view.showProgressBar();
+
+        const context = await DataCaptureContext.forLicenseKey("-- ENTER LICENSE KEY HERE --", {
```
eatsjobs: let's push for the use of DataCaptureContext.sharedInstance also here

docs/sdks/web/add-sdk.md
anthonypenna: Done — dropped the full HTML page. The section now shows only the delta over the CDN example: the two `<script>` tags, the globals table, and a short snippet with the async wrapper. It went from ~113 lines to ~50.

docs/sdks/web/add-sdk.md
anthonypenna: Added as the first thing in the section, a `:::caution Use UMD only as a last resort` admonition covering both points: the UMD bundles cannot be tree-shaken/optimized the way a bundler (or the browser) handles the ES modules, so more code is downloaded and startup is slower — and that ESM is supported by every current browser ([caniuse](https://caniuse.com/?search=ESM)), with [es-module-shims](https://github.com/guybedford/es-module-shims#usage) as the fallback when import map support is the only blocker (which is what the CDN example above already does). @eatsjobs — this covers your reply too.

docs/sdks/web/add-sdk.md
anthonypenna: Done — the snippet now calls `forLicenseKey` once and then uses `DataCaptureContext.sharedInstance` (passed to `view.setContext`), with a comment noting the context is initialized only once and `sharedInstance` is how you reach it from anywhere in the app.

lucatorella (approved):

PR #415, opened 2026-08-05
Search: count whole queries in Algolia analytics, not every keystroke

eugenia-scandit:
The DocSearch widget fires an Algolia request on every keystroke, and each one was counted in Algolia Search Analytics. That inflated the searches total and, worse, the no-result rate: mid-typing prefixes ("l", "li", "lc") frequently return zero hits and were counted as no-result searches, even though they are not real queries and never surface in the Top No Result Queries report. The result was a no-result rate driven by keystroke noise rather than genuine content gaps.

Now the interactive keystroke requests run with analytics:false (results still appear as-you-type; clickAnalytics is untouched so result-click attribution keeps working), and a single analytics:true, count-only request (hitsPerPage:0) is fired per completed query after a 600ms debounce. Queries shorter than 3 chars are skipped, and latestQueryRef is re-checked at fire time so only the final query in a typing burst is recorded. Algolia now counts one search per real query with its true no-result status.

### DIFF
--- a/src/theme/SearchBar/index.js
+++ b/src/theme/SearchBar/index.js
@@ -668,9 +668,11 @@ function DocSearch({ contextualSearch, externalUrlRegex, ...props }) {
             hitComponent={Hit}
             transformSearchClient={transformSearchClient}
             getMissingResultsUrl={({ query }) => {
-              // Most zero-result queries are exact API symbol names that the
-              // main index doesn't contain (data-capture-sdk tree excluded) -
-              // offer the API reference's built-in search as a fallback.
+              // The API reference (data-capture-sdk tree) IS in this index, so
+              // most API-symbol queries do resolve here. This only fires on a
+              // genuine zero-result query - offer the API reference's own
+              // built-in search as a last-resort fallback for symbols/versions
+              // the main index may not cover.
               const sdkFw = (currentFramework || '/sdks/web').replace('/sdks/', '');
               const fw = sdkFrameworkToApi(sdkFw);
               return `https://docs.scandit.com/data-capture-sdk/${fw}/search.html?q=${encodeURIComponent(query)}`;
--- a/src/theme/SearchBar/index.js
+++ b/src/theme/SearchBar/index.js
@@ -545,10 +545,10 @@ function DocSearch({ contextualSearch, externalUrlRegex, ...props }) {
         // Algolia Search Analytics - inflating the searches total and the
         // no-result rate with prefixes that are not real queries. We run the
         // interactive requests with analytics ON but tagged 'as-you-type' (so
-        // click/queryID attribution keeps working and the keystroke prefixes can
-        // be segmented OUT of the whole-query metrics in the dashboard, rather
-        // than lost), and fire ONE untagged count-only request per completed
-        // query via the debounced ping below.
+        // click/queryID attribution keeps working), and fire ONE count-only
+        // request per completed query, tagged 'whole-query', via the debounced
+        // ping below. The dashboard reads two segments: counts + no-result rate
+        // from 'whole-query'; CTR + conversion from 'as-you-type'.
         const buildRequests = (queryOverride, { analyticsPing = false } = {}) =>
           Array.isArray(requests)
             ? requests.map((r) => {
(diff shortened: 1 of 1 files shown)


src/theme/SearchBar/index.js
```diff
@@ -551,6 +562,16 @@ function DocSearch({ contextualSearch, externalUrlRegex, ...props }) {
...
                 }
+                if (analyticsPing) {
+                  // The single counted search per finished query: recorded in
+                  // Search Analytics, but fetch no hits and no queryID - it
+                  // exists only to log the whole query and its no-result status.
+                  params.analytics = true;
+                  params.hitsPerPage = 0;
+                  params.clickAnalytics = false;
+                } else {
+                  params.analytics = false;
```
lucatorella: 🤖 `analytics: false` on the interactive request breaks click-event attribution, contrary to the comment above ("clickAnalytics is untouched so result-click attribution keeps working").

Per Algolia, a click event whose `queryID` came from an `analytics:false` search ["will not be tracked in your analytics, because there will be no search in your analytics to link it back to"](https://support.algolia.com/hc/en-us/articles/13076138034193-What-happens-if-we-set-analytics-as-false-for-a-query-but-set-clickAnalytics-as-true-Will-click-analytics-still-work-and-set-a-queryID-for-searches-so-I-can-send-events). Those events also stop working with NeuralSearch, Dynamic Re-Ranking, Query Categorization and A/B testing, since all of them require the event to be linked to a recorded search.

The counted ping can't carry the clicks instead: it sets `params.clickAnalytics = false` and `params.hitsPerPage = 0`, so it returns neither hits nor a queryID.

This is also the search that the `window.aa = aa` block at the top of the file exists to instrument — that shim was added specifically so click/view events reach Algolia past the CSP. Net effect as written: the searches total becomes accurate while CTR and conversion rate go to zero.

If keystroke suppression is the goal but attribution has to survive, one option is keeping `analytics: true` + `clickAnalytics` on the interactive request and tagging it with `analyticsTags: ['as-you-type']` so the prefixes can be segmented out in the dashboard, rather than excluding them from analytics entirely.

src/theme/SearchBar/index.js
```diff
@@ -581,6 +602,24 @@ function DocSearch({ contextualSearch, externalUrlRegex, ...props }) {
...
+        // after the user pauses typing. Skips <3-char prefixes ("lc") so
+        // mid-typing noise is never counted, and re-checks latestQueryRef at
+        // fire time so only the final query in a burst is recorded. The query
+        // text matches the interactive primary (routed tokens stripped) so its
+        // no-result status reflects what the user actually saw.
+        if (query && query.trim().length >= 3) {
+          if (analyticsCountRef.current) {
+            clearTimeout(analyticsCountRef.current);
+          }
+          const analyticsRequests = buildRequests(strippedQuery, {
```
lucatorella: 🤖 The counted ping ignores the enum-member fallback, so the no-result rate stays inflated — the exact metric this PR is meant to fix.

The interactive path retries with `stripTrailingMember` when the primary returns zero hits and adopts the retry when it finds hits. The ping always fires the un-retried `strippedQuery`. So for a `BarcodeCapture.didScan`-shaped query where the retry succeeds, the user sees results while Algolia records a **zero-result** search.

That also makes the comment above inaccurate: "The query text matches the interactive primary (routed tokens stripped) so its no-result status reflects what the user actually saw." It matches the primary, but the primary is not necessarily what the user saw.

Deriving the ping's query from the resolved `resultPromise` (i.e. from whichever query actually produced the displayed results) would fix the metric and the comment together.

lucatorella (commented):
🤖 Two findings inline — both are about the change trading one broken metric for another.

src/theme/SearchBar/index.js
```diff
@@ -551,6 +563,26 @@ function DocSearch({ contextualSearch, externalUrlRegex, ...props }) {
                     targetTag
                   );
                 }
+                if (analyticsPing) {
+                  // The single counted whole-query record: analytics on, no hits,
+                  // no click attribution, and UNtagged - the clean count once the
+                  // as-you-type prefixes are filtered out in the dashboard.
+                  params.analytics = true;
+                  params.hitsPerPage = 0;
+                  params.clickAnalytics = false;
```
moritzhartmeier, line 573: 🤖 The clean segment is left UNtagged, but Algolia has no way to select "searches with no tag" — so the count this PR produces isn't readable in the dashboard.

The comment here ("UNtagged - the clean count once the as-you-type prefixes are filtered out in the dashboard") and the commit message ("exclude the 'as-you-type' tag to read them") both assume the tag filter supports exclusion. It doesn't. The Analytics API spec for the `tags` parameter says only:

> Tags by which to segment the analytics. You can combine multiple tags with `OR` and `AND`.

No `NOT`, and no "untagged" pseudo-tag. So the segment you *can* select is `as-you-type` — the noisy keystroke one — while the whole-query pings sit in the unfiltered total alongside every keystroke, which is the view the PR set out to fix. Net effect as merged: the default dashboard view is slightly *worse* than before (all keystrokes still counted, plus one extra ping per query), and the accurate number is unreachable.

One line fixes it — give the ping its own tag so it becomes an includable segment:

```js
if (analyticsPing) {
  params.analytics = true;
  params.hitsPerPage = 0;
  params.clickAnalytics = false;
  params.analyticsTags = ["whole-query"];
}
```

Then the dashboard note becomes actionable: filter on `whole-query` for the true search count and no-result rate.

Worth writing into the comment as well: because the ping sets `clickAnalytics: false` and `hitsPerPage: 0`, no click can ever attach to it, so CTR/conversion on the `whole-query` segment is structurally 0 by design — click attribution lives entirely on `as-you-type`. Counts and CTR are readable in two different segments, never one. Without that note it reads as a regression to a future reader.

eugenia-scandit: Tagged the ping whole-query, documented the two-segment read

moritzhartmeier: I see the fix went into https://github.com/Scandit/data-capture-documentation/pull/421 instead of this PR's branch. As the discussion was in here I would prefer to pick the commit here and merge this PR while closing the other one.

moritzhartmeier (approved):

PR #416, opened 2026-08-05
Chore/docs quality gate ci

eugenia-scandit:
Brings the docs quality gate into `main` as focused infrastructure and wires it
into CI so it runs on every pull request.

## What the gate checks (on changed docs only)
- **Vale** — Google Developer Style + Scandit banned terms + a GEO rule
- **cspell** — spelling (advisory / warning)
- **Frontmatter schema** — `docs-schema.yml` conformance (error)
- **Link check** — broken links (error)


lucatorella (commented):
🤖 Two findings inline — both are cases where the gate reports pass without having checked anything, and both are visible in this PR's own green run.

.github/workflows/docs-gate.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: '18'
+          cache: 'yarn'
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Set up Vale
+        run: yarn docs:gate:setup
```
lucatorella, line 47: 🤖 Vale never runs in CI. `setup-vale.cjs` only *detects* Vale, it doesn't install it — and `ubuntu-latest` has no `vale` binary, so the headline check of this PR is permanently skipped while the job stays green.

`scripts/setup-vale.cjs`:

```js
if (!hasVale()) {
  console.log("Vale is not installed. Prose style checks will be skipped until it is.");
  ...
  process.exit(0);
}
```

Then `index.cjs` degrades the same way:

```js
const vale = findVale();
if (vale) findings.push(...runVale(files, vale));
else console.log("docs-gate: Vale not installed — skipping prose style check (…)");
```

This PR's own **passing** `Docs Quality Gate` run ([job 92293717354](https://github.com/Scandit/data-capture-documentation/actions/runs/31002269121/job/92293717354)) confirms it:

```
$ node scripts/setup-vale.cjs
Vale is not installed. Prose style checks will be skipped until it is.
  Windows: winget install errata-ai.Vale
  macOS:   brew install vale
  Linux:   see https://vale.sh/docs/vale-cli/installation/
```

So `.vale.ini`, `styles/Scandit/Banned.yml` and the Google style package are dead weight in CI, and the first gate listed in the PR description ("Vale — Google Developer Style + Scandit banned terms + a GEO rule") enforces nothing. Graceful degradation is right for a developer's laptop; in CI it turns the gate into a no-op that reads as enforcement.

Installing Vale in the workflow (`errata-ai/vale-action`, or a pinned release download before `docs:gate:setup`) plus making the gate hard-fail on a missing Vale when `process.env.CI` is set would make the green mean what it claims.

scripts/docs-gate/index.cjs
```diff
@@ -0,0 +1,117 @@
...
+      const level = a.Severity === "error" ? "error" : "warn";
+      out.push({ file: file.replace(/\\/g, "/"), level, check: `vale:${a.Check}`, msg: `${a.Message} (line ${a.Line})` });
+    }
+  }
+  return out;
+}
+
+function main() {
+  const files = changedDocs();
+  if (files.length === 0) { console.log("docs-gate: no changed docs — nothing to check."); process.exit(0); }
```
lucatorella, line 111: 🤖 An empty file list exits 0, and every git failure that could produce an empty file list is silently swallowed — so "pass" and "we checked nothing" are indistinguishable.

```js
function sh(cmd) {
  return execSync(cmd, { encoding: "utf8", stdio: ["ignore", "pipe", "ignore"] }).trim();
}
```

stderr is discarded, and then every call site catches and drops the error:

```js
try { base = sh("git merge-base origin/main HEAD"); }
catch { try { base = sh("git rev-parse origin/main"); } catch { base = ""; } }
let out = "";
try { out = sh(`git diff --name-only --diff-filter=ACMR ${base} HEAD -- docs`); } catch {}
```

When both lookups fail, `base` is `""` and the interpolated command collapses to `git diff --name-only --diff-filter=ACMR  HEAD -- docs` — a working-tree-vs-HEAD diff, which is empty in a clean CI checkout. Control then reaches this line, prints `no changed docs`, and exits 0. There is no output anywhere that distinguishes that from a genuinely docs-free PR. This PR's own green run ends on exactly that line:

```
docs-gate: no changed docs — nothing to check.
```

which is correct here (no docs changed), and that's the problem — the same green is what a broken ratchet produces.

The base is also hardcoded to `origin/main` while the workflow triggers on `release/**` too:

```yaml
on:
  pull_request:
    branches:
      - main
      - 'release/**'
```

and the fetch step only fetches `main`. For a PR targeting a release branch, `git merge-base origin/main HEAD` lands on the fork point of that release branch, so the "changed docs" set becomes every doc touched on the branch since it forked — the opposite of the ratchet's stated purpose ("never blocks on pre-existing legacy pages").

Deriving the base from `github.event.pull_request.base.ref` (falling back to `origin/main` only outside CI), letting `sh` surface stderr, and failing loudly instead of defaulting `base` to `""` would close both directions.

.github/workflows/docs-gate.yml
lucatorella: 🤖 Verified fixed in `90373fbe`. The workflow now installs a pinned Vale and syncs the styles, and the gate hard-fails in CI when Vale is absent. Confirmed from the run on the new head ([job 92300413688](https://github.com/Scandit/data-capture-documentation/actions/runs/31004322270/job/92300413688)):

```
vale version 3.7.1
$ node scripts/setup-vale.cjs
Syncing Vale style packages (Google)…
 SUCCESS  Synced 1 package(s) to '…/styles'.
Vale styles synced.
```

The `v3.7.1` asset URL resolves (HTTP 200), so the pin is good. One follow-up I've raised separately: the CI hard-fail covers Vale being *missing*, not Vale being present and erroring.

scripts/docs-gate/index.cjs
lucatorella: 🤖 Verified fixed in `90373fbe`. The base now comes from `GITHUB_BASE_REF` so release-branch PRs ratchet against their own target, the workflow fetches `github.base_ref` rather than hardcoded `main`, `sh()` inherits stderr instead of discarding it, and an unresolvable base `exit(2)`s in CI instead of collapsing to an empty diff.

One note on the remaining `git diff` call: `try { out = sh(...) } catch {}` still swallows a failure there, which lands back on the `files.length === 0` early exit. Much narrower now that the base is validated, so not worth blocking on — but the gate still cannot be demonstrated end-to-end on this PR, since it changes no docs and the run still ends at `docs-gate: no changed docs — nothing to check.` A throwaway commit touching one doc with a deliberate banned term (`simply`) would prove the whole chain, including that Vale alerts actually reach the error count.

scripts/docs-gate/index.cjs
```diff
@@ -0,0 +1,149 @@
...
+}
+
+function runVale(files, bin) {
+  const out = [];
+  let json;
+  try {
+    const raw = execFileSync(bin, ["--output=JSON", ...files], { cwd: ROOT, encoding: "utf8", stdio: ["ignore", "pipe", "pipe"] });
+    json = JSON.parse(raw);
+  } catch (e) {
+    try { json = JSON.parse(e.stdout || "{}"); } catch { return out; }
```
lucatorella: 🤖 The fix closes "Vale missing → green" but leaves "Vale present and failing → green", which is the same swallow-and-return-empty shape.

```js
try {
  const raw = execFileSync(bin, ["--output=JSON", ...files], { … });
  json = JSON.parse(raw);
} catch (e) {
  try { json = JSON.parse(e.stdout || "{}"); } catch { return out; }
}
```

When Vale exits non-zero for a reason other than "found alerts" — a config error, an unparseable rule, a `StylesPath` that isn't there — it writes a diagnostic to stderr and nothing to stdout. `e.stdout` is then `""`, `JSON.parse("{}")` succeeds, `Object.entries({})` iterates nothing, and the function returns an empty findings array. Zero Vale errors, gate exits 0, job green.

The new CI guard doesn't cover this, because it keys off `findVale()`:

```js
const vale = findVale();
if (vale) {
  findings.push(...runVale(files, vale));
} else if (process.env.CI) { /* hard error */ }
```

and `findVale()` only probes `vale -v`. A binary that runs `-v` fine but errors on the lint invocation takes the `if (vale)` branch and contributes nothing.

This is reachable without anyone doing something exotic: `.vale.ini` declares `BasedOnStyles = Google, Scandit` with `Packages = Google`, and the `Google` package only exists in `styles/` because the separate `yarn docs:gate:setup` step ran first. Reorder or drop that step, restore a cache without it, or land an invalid regex in `styles/Scandit/Banned.yml`, and every prose check silently stops producing findings while the gate keeps saying pass — exactly the state `90373fbe` set out to eliminate.

Distinguishing the two exit paths would fix it: treat a non-zero exit whose stdout isn't valid JSON as a gate error (echo Vale's stderr into the message) rather than as "no alerts".

lucatorella (commented):
🤖 Re-reviewed at `90373fbe` — both earlier findings verified fixed (replies in their threads). One instance of the same defect class is still open.

scripts/docs-gate/index.cjs
lucatorella: 🤖 Verified fixed in `809314f0`. A non-zero exit whose stdout isn't valid JSON is now a blocking gate error carrying Vale's stderr, instead of being read as "no alerts":

```js
try {
  json = JSON.parse(stdout);
} catch {
  const stderr = ((e.stderr || "").toString().trim() || (e.message || "").trim()).split(/\r?\n/).slice(0, 5).join(" ");
  return [{ file: ".vale.ini", level: "error", check: "vale", msg: `Vale failed to run (not an alert): ${stderr}` }];
}
```

The alerts path is preserved (non-zero + JSON stdout still parses), and a `JSON.parse` failure on the *success* path lands in the same branch with the SyntaxError as the message, so that's loud too. The run on the new head is green with no false `Vale failed to run`, and `vale sync` reports `Vale styles synced.`

Two caveats, neither blocking:

1. The new branch is still unexercised by CI — this PR changes no docs, so the run ends at `docs-gate: no changed docs — nothing to check.` and `runVale` is never called. The end-to-end proof is still outstanding.
2. `runCspell` has the same shape one function up:

```js
catch (e) {
  const lines = ((e.stdout || "") + (e.stderr || "")).split(/\r?\n/).filter((l) => / - Unknown word /.test(l));
  return lines.map(…);
}
```

A cspell that fails to run (bad `cspell.json`, invalid glob) yields no lines matching `/ - Unknown word /` and returns `[]` — silently zero findings. It doesn't change the gate's verdict today because cspell is `level: "warn"`, but the comment there says "Promote to `error` once the dictionary matures", and at that point this becomes the same blocking-relevant hole you just closed for Vale.

lucatorella (approved):

PR #424, opened 2026-08-13
ci: add GitHub Pages PR preview deploys

eatsjobs:
Adds a preview-link workflow for docs PRs.

## Verification

- Open → deploy: sticky comment posted, preview loads at `https://scandit.github.io/data-capture-documentation/pr-preview/pr-424/`.
- Close → remove: confirmed the preview URL 404s after closing, build job correctly skipped.
- Reopen → redeploy: confirmed the URL is back to 200 after reopening.
- Isolation between concurrent previews: #425 (kept open as evidence, not merged) deployed its own preview at `pr-preview/pr-425/` without disturbing this PR's preview, and vice versa.

## Review

Addressed the size/1GB-cap concern, SHA-pinning, QR code, and job-split feedback from review — see inline replies and https://github.com/Scandit/data-capture-documentation/pull/424#issuecomment-5283917564 for details.

---

@moritzhartmeier — resolved, Pages is now serving from `gh-pages` and the preview link above works.


moritzhartmeier (commented):
A few smaller points beyond the size comment above. None are blockers on their own.

moritzhartmeier (commented):
Built the branch locally with `preview_build=true` to check this rather than take the numbers on faith. It all holds up.

**Size: 304.3 MB → 77 MB.** Your measurement matches mine (the 71.5 MB deployed vs my 77 MB is just `du` block-size vs blob bytes):

```
38 MB  sdks        18 MB  assets      16 MB  img  (was 186 MB)
 2 MB  llms-full  1.5 MB  hosted       0 gif/mp4 files remaining
```

**The media offload is complete — no dangling references.** I grepped the built HTML and JS for local `/img/**.gif|mp4` paths that no longer exist: zero hits. All 30 heavy-media references were rewritten to `docs.scandit.com`, including the ones inside imported partials (`_batch-scanning.mdx`, `_single-scanning.mdx`, the `ai-powered-scanning` mp4s) — that was the case I'd have expected to slip, so good that the remark plugin reaches partials. I also checked the reference shapes across `docs/`: every gif/mp4 is a JSX `` or `<ReactPlayer url>` with an absolute `/img/` path, and there are no markdown-syntax `` refs at all, so the plugin's narrow matcher covers the real corpus rather than happening to work on the pages I looked at.

**`.nojekyll` survives the artifact round-trip.** I half-expected `upload-artifact@v4` to drop it (v4.4+ excludes hidden files by default) — it doesn't; it's present in the deployed `pr-preview/pr-424/`. Non-issue, just noting it since it would have been a quiet breakage.

**`onBrokenLinks: "warn"` is contained** — my build emitted exactly the one warning you described (`/sdks/web/release-notes/` → `/7.6.14/…`), not a general loosening.

One small thing inline. Otherwise this looks good to me.

---

**Answer to your open question on the build-docs dedup:** `Build Docusaurus Site` *is* a required status check on `main` —

```
branch protection → required_status_checks.contexts = ["Build Docusaurus Site"]
```

so your caution was right. If anyone picks that follow-up up, the dedup has to keep that exact job name or update the ruleset in the same change, otherwise every PR blocks on a check that no longer reports.

moritzhartmeier (approved):

PR #428, opened 2026-08-20
 fix(search): restore the API reference to search and stop docs-version renames from breaking it

eugenia-scandit:
## What was broken

Releasing 8.5.3 renamed the docs version, so `docusaurus_tag` on every page at
the site root changed from `docs-default-current` to `docs-default-8.5.3` —
while not a single URL changed.

Two consequences, both silent:

1. The API reference lives in the same index but is not a Docusaurus version, so
   it carried the tag `docs-default-current`. While that tag matched the guides'
   tag, it was found alongside them for free. After the rename ~3,200 pages left
   every result. A query with 106 matching pages in the index returned 86.
2. `buildVersionTagByMajor` let `current` win its major unconditionally. With
   8.6.0-beta as `current` and 8.5.3 as `lastVersion`, typing "v8" sent readers
   at the unreleased beta's tag — and it still returned results, so nothing
   looked wrong.

The monitoring could only report "too few records".

## What this changes

- `docusaurus.config.ts` is the only place versions are declared and tags are
  built. `DOCS_LAST_VERSION` is declared once beside `docsVersions` (it used to
  be restated in the preset), and `docVersionTag` is the only constructor of
  `docs-default-*` strings.
- `buildVersionTagByMajor` follows `lastVersion` and skips unreleased versions.
- The API reference is mapped per version instead of bolted on: each docs
  version gets the API tree it belongs with — `7.6.14 → api-reference-7.6`,
  `6.28.11 → api-reference-6.28`, and the version served at the root → the
  unversioned tree. A reader on 6.28.11 no longer gets 8.x API pages.
- Fixes a regression introduced in this same PR: hoisting `DOCS_LAST_VERSION`
  left `scripts/update-version.py` reading the `lastVersion:` line, so the next
  release would have aborted with "Already in production state". Verified by
  running a real release end to end.

## Why the mapping mirrors the site rather than assuming symmetry

The sitemap contains zero `/data-capture-sdk/` URLs, so the crawler can only
reach that tree by following links. The 7.6.14 guides link to
`/7.6/data-capture-sdk/…`, but the guides served at the root link to the
**unversioned** `/data-capture-sdk/…` — nothing anywhere links to `/8.5/`.
Pointing the served version at `api-reference-8.5` would have pointed at a tree
the crawler never reaches, dropping the current API reference out of search a
second time.

## Preventing a repeat

- `yarn verify:search-tags` (in CI) compares the build against the live index:
  the tag a typed major routes to must be the tag the site serves; the build must
  emit every tag the config names; and content under a tag search cannot reach is
  an error. Crawler lag warns instead of failing.
- `yarn test:search-facets` (in CI) — 7 assertions on what the widget actually
  sends to Algolia, including that a legacy version gets *its* API reference and
  not the current one.

Both reproduce both bugs against the pre-fix config.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -30,16 +30,17 @@ jobs:
       - name: Build documentation
         run: yarn build
 
-      # Compares build/search-tags.json (what this build's widget will filter on)
-      # against the live Algolia index. Catches a docs-version rename quietly
-      # dropping content out of search - see scripts/verify-search-tags.cjs.
-      # Not --strict on PRs: a crawl still catching up after a deploy warns
-      # rather than blocking unrelated work.
       # Asserts what the widget actually sends to Algolia: the API reference is
       # reachable on the served version and absent on frozen legacy versions.
       # Both failure modes are silent in production.
       - name: Test search facet filters
         run: yarn test:search-facets
 
+      # Compares build/search-tags.json (what this build's widget will filter on)
+      # against the live Algolia index. Catches a docs-version rename quietly
+      # dropping content out of search - see scripts/verify-search-tags.cjs.
+      # Not --strict on PRs: a crawl still catching up after a deploy, a release
+      # build that renames the served version, or an Algolia outage all warn
+      # rather than blocking unrelated work.
       - name: Verify search tags
         run: yarn verify:search-tags
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -29,3 +29,11 @@ jobs:
 
       - name: Build documentation
         run: yarn build
+
+      # Compares build/search-tags.json (what this build's widget will filter on)
+      # against the live Algolia index. Catches a docs-version rename quietly
+      # dropping content out of search - see scripts/verify-search-tags.cjs.
+      # Not --strict on PRs: a crawl still catching up after a deploy warns
+      # rather than blocking unrelated work.
+      - name: Verify search tags
+        run: yarn verify:search-tags
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -35,5 +35,11 @@ jobs:
       # dropping content out of search - see scripts/verify-search-tags.cjs.
       # Not --strict on PRs: a crawl still catching up after a deploy warns
       # rather than blocking unrelated work.
+      # Asserts what the widget actually sends to Algolia: the API reference is
+      # reachable on the served version and absent on frozen legacy versions.
+      # Both failure modes are silent in production.
+      - name: Test search facet filters
+        run: yarn test:search-facets
+
       - name: Verify search tags
         run: yarn verify:search-tags
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -42,5 +42,11 @@ jobs:
       # Not --strict on PRs: a crawl still catching up after a deploy, a release
       # build that renames the served version, or an Algolia outage all warn
       # rather than blocking unrelated work.
+      # --strict on main only. On a PR a crawl still catching up, a release build
+      # that renames the served version, or an Algolia outage must not block
+      # unrelated work; on main those same conditions are real and someone is
+      # watching. Without this the script's repeated "--strict still fails"
+      # promise was empty - THIN, transport errors and Algolia-key drift could
+      # not fail any job.
       - name: Verify search tags
-        run: yarn verify:search-tags
+        run: yarn verify:search-tags${{ github.ref == 'refs/heads/main' && ' --strict' || '' }}
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -8,6 +8,24 @@ on:
   push:
     branches:
       - main
+  # The post-crawl --strict run the search-tag gate is designed around.
+  #
+  # Every other trigger fires BEFORE the Algolia crawler has seen the deploy, so
+  # "the index has not caught up" is indistinguishable from "the index is wrong"
+  # and the gate can only warn. Once a day, hours after any merge, that ambiguity
+  # is gone: a tag still holding nothing is really holding nothing. No extra
+  # --strict wiring is needed - a scheduled run's github.ref IS refs/heads/main,
+  # so the expression on the verify step already picks it up.
+  #
+  # Without this, every condition the gate downgrades was downgraded everywhere
+  # and permanently: the release build warns, the push to main warns for the same
+  # pre-crawl reason, and an unreachable API reference would only surface on some
+  # later unrelated PR - in a green log nobody reads.
+  schedule:
+    - cron: "0 6 * * *"
+  # And so it can be run by hand right after a deploy, which is what the gate's
+  # own "re-run after the crawl" messages ask for.
+  workflow_dispatch:
 
 jobs:
   build:
@@ -39,14 +57,16 @@ jobs:
       # Compares build/search-tags.json (what this build's widget will filter on)
       # against the live Algolia index. Catches a docs-version rename quietly
       # dropping content out of search - see scripts/verify-search-tags.cjs.
-      # Not --strict on PRs: a crawl still catching up after a deploy, a release
-      # build that renames the served version, or an Algolia outage all warn
-      # rather than blocking unrelated work.
-      # --strict on main only. On a PR a crawl still catching up, a release build
-      # that renames the served version, or an Algolia outage must not block
-      # unrelated work; on main those same conditions are real and someone is
-      # watching. Without this the script's repeated "--strict still fails"
-      # promise was empty - THIN, transport errors and Algolia-key drift could
-      # not fail any job.
+      # --strict everywhere except pull requests. On a PR, a crawl still catching
+      # up after a deploy, a release build that renames the served version, or an
+      # Algolia outage must not block unrelated work; on main and on the daily
+      # schedule those same conditions are real and someone is watching. Without
+      # any --strict run the script's escape hatches were unconditional, so THIN,
+      # transport errors and Algolia-key drift could not fail any job.
+      #
+      # Safe on push-to-main even though that run is pre-crawl: the pending-release
+      # state is a fact about the index, not a severity preference, so --strict
+      # deliberately does NOT override it (see releasePending in the script). A
+      # release merge warns; it does not go red.
       - name: Verify search tags
         run: yarn verify:search-tags${{ github.ref == 'refs/heads/main' && ' --strict' || '' }}
--- a/docusaurus.config.ts
+++ b/docusaurus.config.ts
@@ -129,6 +129,11 @@ const llmsIgnoreFiles: string[] = [
 // requires lastVersion to be one of the included versions, so previews follow it.
 const DOCS_LAST_VERSION = "8.5.3";
 
+// The version actually served at the root of THIS build. Preview builds only
+// build `current`, so routing a major at a frozen tag would point at pages the
+// build does not contain. Declared once: it was restated at four call sites.
+const effectiveLastVersion = isPreviewBuild ? "current" : DOCS_LAST_VERSION;
+
 const docsVersions: Record<
   string,
   {
@@ -173,11 +178,17 @@ const docVersionTag = (versionName: string): string =>
  * the index and fails the build if the crawler is retagged.
  *
  * Why this exists: while `lastVersion` was "current" the API reference happened
- * to share the guides' tag and rode along for free. Releasing 8.5.3 renamed the
- * guides' version, the shared tag stopped being shared, and ~3,200 API reference
- * pages silently dropped out of every search. Nothing failed - results just got
- * quietly worse. Listing the tag explicitly is what makes it survive the next
- * rename.
+ * to share the guides' tag and rode along for free. That stopped at e92c1b16
+ * (Release 8.6.0-beta.1), which set `lastVersion: "8.5.2"` and made `current`
+ * the unreleased beta - the guides at the root stopped emitting
+ * `docs-default-current`, the shared tag stopped being shared, and ~3,200 API
+ * reference pages silently dropped out of every search. Nothing failed; results
+ * just got quietly worse. 8.5.3 was a later patch bump, not the cause.
+ *
+ * And it is a CYCLE, not a one-off: the root-served tag moves on the
+ * production -> beta transition, on every patch during the beta window, and
+ * again on beta -> production. Listing the tag explicitly is what makes it
+ * survive each turn of that cycle.
  */
 /**
  * The API reference is published per major.minor line, at
@@ -208,11 +219,10 @@ const apiReferenceTag = (versionNumber: string): string =>
  * Both sides read the version out of what they already have: this file out of
  * docsVersions, the crawler out of the URL. Neither hard-codes one, which is the
  * point - tagging the API reference with a docs version NAME is what broke
- * search when releasing 8.5.3 renamed it.
+ * search when the 8.6.0-beta.1 release moved the root-served tag (e92c1b16).
  */
 function buildApiReferenceTags(
   versions: Record<string, { label?: string }>,
-  legacyTag: string,
   lastVersion: string,
 ): Record<string, string[]> {
   const out: Record<string, string[]> = {};
@@ -223,11 +233,6 @@ function buildApiReferenceTags(
     const tags = [
       servedAtRoot ? "api-reference-latest" : apiReferenceTag(number),
     ];
-    // Migration: until the crawler is retagged, the whole API reference still
-    // carries one legacy tag. Keep it on the served version so search works
-    // whichever change lands first; `yarn verify:search-tags` says when the new
-    // tags are populated and this line can go.
-    if (name === lastVersion) tags.push(legacyTag);
     out[docVersionTag(name)] = tags;
   }
   return out;
@@ -295,7 +300,7 @@ const versionTagByMajor = buildVersionTagByMajor(
   docsVersions,
   // Previews only build `current`, so route every major there rather than at
   // frozen tags whose pages that build does not contain.
-  isPreviewBuild ? "current" : DOCS_LAST_VERSION,
+  effectiveLastVersion,
 );
 
 /**
@@ -309,7 +314,7 @@ function searchTagsManifestPlugin() {
     async postBuild({ outDir }: { outDir: string }) {
       const { writeFile } = await import("fs/promises");
       const { join } = await import("path");
-      const lastVersion = isPreviewBuild ? "current" : DOCS_LAST_VERSION;
+      const lastVersion = effectiveLastVersion;
       await writeFile(
         join(outDir, "search-tags.json"),
         JSON.stringify(
@@ -321,7 +326,6 @@ function searchTagsManifestPlugin() {
             lastVersionTag: docVersionTag(lastVersion),
             apiReferenceTagsByVersionTag: buildApiReferenceTags(
               docsVersions,
-              docVersionTag("current"),
               lastVersion,
             ),
             versionTagByMajor,
@@ -351,8 +355,7 @@ const config: Config = {
     // reader on 6.28.11 finds the 6.28 API and never the 8.x one.
     apiReferenceTagsByVersionTag: buildApiReferenceTags(
       docsVersions,
-      docVersionTag("current"),
-      isPreviewBuild ? "current" : DOCS_LAST_VERSION,
+      effectiveLastVersion,
     ),
   },
 
@@ -654,7 +657,7 @@ const config: Config = {
           includeCurrentVersion: true,
           // See DOCS_LAST_VERSION above - declared once, next to docsVersions,
           // so the search tag derivation and the docs plugin cannot disagree.
-          lastVersion: isPreviewBuild ? "current" : DOCS_LAST_VERSION,
+          lastVersion: effectiveLastVersion,
           versions: docsVersions,
         },
         blog: false,
--- a/docusaurus.config.ts
+++ b/docusaurus.config.ts
@@ -179,7 +179,34 @@ const docVersionTag = (versionName: string): string =>
  * quietly worse. Listing the tag explicitly is what makes it survive the next
  * rename.
  */
-const ALWAYS_ON_SEARCH_TAGS: string[] = [docVersionTag("current")];
+const ALWAYS_ON_SEARCH_TAGS: string[] = [
+  // Target state: a stable tag that is not a version name, so no release can
+  // ever rename it out from under the widget. Set this in the Algolia crawler's
+  // /data-capture-sdk/** action (the `docusaurus_tag` field of the record it
+  // returns). Until that crawl lands the tag holds nothing, which the gate
+  // reports as a pending migration rather than a failure.
+  "api-reference",
+  // Current state, kept so search keeps working whichever change lands first.
+  // Drop this entry once `yarn verify:search-tags` shows api-reference
+  // populated - it is also the 8.6.0 beta's own tag, which is why it is a bad
+  // long-term home for the API reference.
+  docVersionTag("current"),
+];
+
+/**
+ * The versions on which always-on content is in scope.
+ *
+ * The API reference documents the current SDK, so a reader on 7.6.14 or 6.28.11
+ * must not get it: measured on a legacy page, pulling it in took one query from
+ * 53 hits to 73, and the 20 extra were 8.x API pages that do not apply there.
+ * Before the 8.5.3 rename this was handled by accident - the API reference
+ * shared the served version's tag, so it appeared there and nowhere else. This
+ * makes that scoping explicit instead of incidental.
+ */
+const ALWAYS_ON_SCOPE_VERSIONS = (lastVersion: string): string[] => [
+  docVersionTag(lastVersion),
+  docVersionTag("current"),
+];
 
 /**
  * Map a major version typed in a query ("v7", "sdk 6") to the tag of the version
@@ -268,6 +295,7 @@ function searchTagsManifestPlugin() {
             // The tag every page at the site root emits.
             lastVersionTag: docVersionTag(lastVersion),
             alwaysOnTags: ALWAYS_ON_SEARCH_TAGS,
+            alwaysOnScopeTags: ALWAYS_ON_SCOPE_VERSIONS(lastVersion),
             versionTagByMajor,
           },
           null,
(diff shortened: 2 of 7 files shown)


src/theme/SearchBar/index.js
```diff
@@ -139,18 +172,28 @@ function versionTagInQuery(query, versionTagByMajor) {
...
       return f.startsWith("docusaurus_tag:docs-")
         ? `docusaurus_tag:${targetTag}`
         : f;
     }
-    return Array.isArray(f) ? f.map(swap) : f;
+    if (!Array.isArray(f)) return f;
+    // Swap the whole API-reference set for the target version's, keeping
+    // docusaurus_tag:default and anything else in the OR group untouched.
+    const rest = f.filter((t) => !String(t).startsWith(API_TAG_PREFIX));
+    return [...rest.map(swap), ...targetApi];
```
moritzhartmeier: **This ANDs the API-reference tag against the whole query.** `targetApi` is appended at every array level, and `facetFilters` is itself an array, so the top level gets it too. Actual output for a reader on 8.5.3 who types "v7":

```
["language:en",
 ["…:default", "…:docs-default-7.6.14", "…:docs-default-7.6.14", "…:api-reference-7.6"],
 "docusaurus_tag:api-reference-7.6"]     ← top level = AND
```

Three top-level entries. That last one means **only 7.6 API reference pages can match — every guide page is filtered out**, the exact opposite of "typing a version moves the guides AND the API reference together." Lines 141-142 above already say *"appending at the top level would AND it against the page's own tag and match nothing at all."*

Append `targetApi` only when recursing into the OR group, not at whatever level `swap` is called on.

Also visible above: `docs-default-7.6.14` appears twice, because the legacy `docs-default-current` entry also matches `docusaurus_tag:docs-` and gets swapped too — a symptom of the shim; see the note on `docusaurus.config.ts:230`.

scripts/test-search-facets.cjs
```diff
@@ -0,0 +1,113 @@
...
+
+check("typing a version moves the guides AND the API reference together", () => {
+  const filters = ["language:en", [DEFAULT, SERVED, APILATEST]];
+  const out = rewriteVersionTag(filters, "docs-default-7.6.14", MAP);
+  const tags = tagsOf(out);
+  assert.ok(tags.includes(LEGACY), "page tag must be swapped");
+  assert.ok(tags.includes(API76), "API reference must follow to 7.6");
+  assert.ok(!tags.includes(APILATEST), "the current API reference must be dropped");
+  assert.ok(tags.includes(DEFAULT), "the OR group must not collapse");
+});
```
moritzhartmeier: This is the test that should have caught the bug in `rewriteVersionTag`, and it structurally can't: `tagsOf = (filters) => filters.find(Array.isArray)` only inspects the nested OR group, so a stray top-level entry is invisible to every assertion here.

Compare line 89, which *does* assert `out.length === 2` with the comment "a top-level entry would AND and match nothing." Add the same assertion here and the bug fails the suite immediately.

scripts/verify-search-tags.cjs
```diff
@@ -0,0 +1,333 @@
...
+  // build: the repo can name the crawler's target tag before the crawler emits
+  // it, or the other way round. Either order must keep CI green as long as one
+  // always-on tag still carries the content.
+  // Flatten the per-version API-reference map: every tag it names must exist.
+  const apiMap = manifest.apiReferenceTagsByVersionTag || {};
+  const alwaysOn = [...new Set(Object.values(apiMap).flat())];
+  // API-reference tags are reported by the migration check above, so they are
+  // excluded here rather than counted twice.
+  const apiTags = new Set(alwaysOn);
+  const missing = [manifest.defaultTag, manifest.lastVersionTag, ...routable]
```
moritzhartmeier: **This fails CI on every release PR — the change the gate exists to protect.** I simulated the next release (8.5.3 -> 8.5.4, exactly what `update-version.py` writes) against the live index:

```
FAIL: indexed content the search widget cannot reach.
  559 pages tagged "docs-default-8.5.3" are in the index but filtered out of every query.

FAIL: the widget filters on tags that hold nothing at all.
  "docs-default-8.5.4" -> 0 records
  "docs-default-8.5.4" -> 0 records
```

Both are "the live index hasn't caught up with this build" — the condition the PR body says should warn rather than block. But `missing` here and `unreachable` at line 260 both set `failed = true` unconditionally, while `thin` — the *less* severe case — is the only one that respects `--strict`. That's inverted for the release flow, and a release PR has no path to green until after deploy plus crawl.

Needs to recognise "lastVersionTag is new in this build" explicitly: warn instead of fail, and don't count the outgoing version's tag as orphaned on the build that renames it.

(Minor, same block: `docs-default-8.5.4` prints twice — `missing` isn't deduped, since `lastVersionTag` is also in `routable`.)

docusaurus.config.ts
```diff
@@ -138,13 +158,94 @@ const docsVersions: Record<
...
+    if (!number) continue;
+    const servedAtRoot = name === lastVersion || name === "current";
+    const tags = [
+      servedAtRoot ? "api-reference-latest" : apiReferenceTag(number),
+    ];
+    // Migration: until the crawler is retagged, the whole API reference still
+    // carries one legacy tag. Keep it on the served version so search works
+    // whichever change lands first; `yarn verify:search-tags` says when the new
+    // tags are populated and this line can go.
+    if (name === lastVersion) tags.push(legacyTag);
```
moritzhartmeier: **This shim is already obsolete, it hides an orphan, and it's a latent leak.**

Its premise is that the whole API reference still carries one legacy tag. The live index says the retagging is done: `api-reference-latest` 3,407 pages, `docs-default-current` **1** page — a stale `/sdks/linux/barcode-capture/get-started/` record.

- Its only current effect is injecting that one orphan into every query on the served docs.
- By naming it an always-on tag, it makes the gate print `1  reachable  docs-default-current` instead of flagging it for purging — hiding exactly the drift the gate exists to surface.
- The gate can't tell you it's removable either: `pendingAlwaysOn` (verify-search-tags.cjs:281) only fires when a tag has *zero* records, and this one has one. The comment's "verify:search-tags says when this line can go" isn't implemented.

And the landmine: the sitemap lists **616 `/next/` URLs** (the 8.6.0 beta), currently excluded by the crawler — which is why only that stale page carries the tag. If that exclusion ever changes, this line silently pulls 616 unreleased-beta pages into normal search results on the served docs. That's the "too wide" failure you fixed for legacy versions, armed for the served version.

It also degenerates after promotion: with `lastVersion: "current"` the shim pushes `docs-default-current` as an API tag for the version whose own tag *is* `docs-default-current`, and the dedupe in `withApiReferenceTags` drops it. Incoherent in both states — just delete it.

scripts/verify-search-tags.cjs
```diff
@@ -0,0 +1,333 @@
+#!/usr/bin/env node
+/**
+ * Search-tag gate.
+ *
+ * The search widget filters every query on `docusaurus_tag`, and Docusaurus
+ * builds that tag from the version NAME. So renaming a docs version - which is
+ * what every release does - changes what search can reach, without changing a
+ * single URL and without breaking anything loudly. In August 2026 releasing
+ * 8.5.3 renamed the guides from `docs-default-current` to `docs-default-8.5.3`;
```
moritzhartmeier: The stated cause is wrong here, in the `docusaurus.config.ts` SSOT block, and in three commit messages. Releasing 8.5.3 didn't rename the version — `e92c1b16` Release 8.6.0-beta.1 did, by setting `lastVersion: "8.5.2"` and making `current` the unreleased beta. 8.5.3 was a later patch bump.

Worth fixing because the real pattern is more aggressive than the one described: the root-served tag also moves on the production->beta transition and again on beta->production, so this is a cycle rather than a one-time event. A reader of this docstring wouldn't expect that.

scripts/verify-search-tags.cjs
```diff
@@ -0,0 +1,333 @@
...
+      `  Usually the crawler catching up after a deploy; re-run once it has.`,
+    );
+    if (strict) failed = true;
+  }
+
+  if (!failed) console.log(`\nOK: every tag with content is reachable.\n`);
+  process.exit(failed ? 1 : 0);
+}
+
+main().catch((err) => {
```
moritzhartmeier: Network failure is a hard CI failure. An Algolia outage, a rate-limit, or a sandboxed runner blocks merging unrelated docs PRs. Transport errors should warn unless `--strict`; run `--strict` on main or on a schedule where a real failure gets attention.

scripts/verify-search-tags.cjs
```diff
@@ -0,0 +1,333 @@
...
+
+/**
+ * Read the docusaurus_tag values the build actually emits.
+ *
+ * The manifest states what this build BELIEVES it serves. This reads what it
+ * really wrote into the HTML. They are the same fact from two directions, which
+ * is the only way a derived value gets checked - the drift that caused this gate
+ * to exist was a config that was internally consistent and still wrong.
+ */
+function tagsEmittedByBuild(buildDir, expected, fileCap = 4000) {
```
moritzhartmeier: `fileCap = 4000` with an arbitrary `stack.pop()` DFS order can produce a false "config drift" failure: if the cap is hit before all wanted tags are found, the output is `FAIL: the config names tags this build never emits`, blaming the config for a truncated walk.

The build is currently ~2,250 pages (676 current + 1,573 versioned; the API reference isn't in this repo's `static/`), so roughly two frozen versions of headroom. Report inconclusive when the cap is reached, or locate one file per wanted tag deterministically.

scripts/test-search-facets.cjs
```diff
@@ -0,0 +1,113 @@
...
+    } else if (src[i] === "}") {
+      depth -= 1;
+      if (started && depth === 0) return src.slice(start, i + 1);
+    }
+  }
+  throw new Error(`could not read ${name}`);
+}
+
+const EMPTY_TAG_LIST = [];
+const API_TAG_PREFIX = "docusaurus_tag:api-reference-";
```
moritzhartmeier: These two constants are re-declared locally, so the `eval`'d functions close over the test's copies rather than the module's. The docstring above says the functions are read from the shipped module because "a copy would keep passing after the real one changed" — and `API_TAG_PREFIX` is precisely the constant whose change would go unnoticed here.

docusaurus.config.ts
```diff
@@ -154,27 +255,85 @@ function buildVersionTagByMajor(
...
   }
   return out;
 }
 
-const versionTagByMajor = buildVersionTagByMajor(docsVersions);
+const versionTagByMajor = buildVersionTagByMajor(
+  docsVersions,
+  // Previews only build `current`, so route every major there rather than at
+  // frozen tags whose pages that build does not contain.
+  isPreviewBuild ? "current" : DOCS_LAST_VERSION,
```
moritzhartmeier: `isPreviewBuild ? "current" : DOCS_LAST_VERSION` is restated four times (here, 312, 355, 657). One `effectiveLastVersion` constant beside `DOCS_LAST_VERSION`.

scripts/verify-search-tags.cjs
```diff
@@ -0,0 +1,333 @@
...
+
+  console.log(`\nIndex "${ALGOLIA.indexName}" - distinct pages per docusaurus_tag`);
+  console.log(`Site serves: ${manifest.lastVersionTag} (lastVersion ${manifest.lastVersion})\n`);
+  for (const [tag, n] of rows) {
+    console.log(
+      `  ${String(n).padStart(6)}  ${
+        reachable.has(tag)
+          ? "reachable  "
+          : routable.has(tag)
+            ? "typed-only "
```
moritzhartmeier: `api-reference-7.6` / `api-reference-6.28` print as `typed-only`, but they're contextually reachable from those versions' pages — the comment just above says as much. Mislabel in the operator-facing report.

scripts/verify-search-tags.cjs
```diff
@@ -0,0 +1,333 @@
...
+ * facet count is several times the page count and reads far healthier than the
+ * widget's own result count. nbHits under a facet filter applies the index's
+ * `distinct` setting and is what a reader actually sees.
+ */
+async function pagesByTag() {
+  const { facets } = await algolia({
+    query: "",
+    hitsPerPage: 0,
+    facets: ["docusaurus_tag"],
+    maxValuesPerFacet: 100,
```
moritzhartmeier: `maxValuesPerFacet: 100` — above 100 distinct tags, absent ones read as `undefined` and land in `missing` as a false FAIL. Only 10 today, so just a note. Same block: `pagesByTag` fans out one query per tag via `Promise.all` with no concurrency cap.

.github/workflows/build-docs.yml
```diff
@@ -29,3 +29,17 @@ jobs:
 
       - name: Build documentation
         run: yarn build
+
+      # Compares build/search-tags.json (what this build's widget will filter on)
```
moritzhartmeier: One comment block describing both steps, in the reverse order of the steps below it — the `--strict`/crawl-lag paragraph belongs above "Verify search tags", not above the facet test.

moritzhartmeier (changes requested):
**Two blocking bugs. Everything else is smaller.**

Verified by execution: ran both new scripts, queried the live Algolia index, simulated a release bump through the gate, dry-ran both `update-version.py` transitions on this branch, and typechecked (`tsc` clean). All 5 CI checks pass and the 7 facet tests pass — the two bugs below are ones the tests are shaped not to catch.

**What holds up:** the `buildVersionTagByMajor` priority fix is correct and is a genuine independent bug — confirmed major 8 now resolves to `docs-default-8.5.3` rather than the unreleased beta. The `update-version.py` fix works: I ran `8.5.3 -> 8.5.4` and then the `8.6.0` promotion, and the constant, the `docsVersions` key and `versions.json` all land correctly, ending at `DOCS_LAST_VERSION = "current"`. The per-version API mapping is also cycle-stable — it still derives correctly after promotion, when `current` becomes the served version again. And checking a build artifact against the live index rather than re-deriving config against itself is the right instinct.

### Blocking

**1. `rewriteVersionTag` ANDs the API-reference tag against the whole query, so typing "v7" returns zero guides.** Reproduced; detail inline. Lines 141-142 of the same file warn against exactly this.

**2. The gate hard-fails on every release PR** — the change it exists to protect. Simulated 8.5.3 -> 8.5.4 against the live index: two blocking FAILs, neither gated by `--strict`. Detail inline.

### One factual correction

Releasing 8.5.3 didn't rename the docs version. That happened at **`e92c1b16` Release 8.6.0-beta.1 (#423)**, which set `lastVersion: "8.5.2"` and made `current` the unreleased beta; 8.5.3 was a later patch bump. The wrong cause is stated in the `docusaurus.config.ts` SSOT block, in the gate's docstring, and in three commit messages, so it's worth correcting where it will be read later.

It also isn't a one-off: the root-served tag moves on the production->beta transition, on patches during the beta window, and again on beta->production. That's a cycle, which strengthens the case for *something* here — but it means the migration shim in particular needs to go rather than be carried forward (see inline).

scripts/verify-search-tags.cjs
moritzhartmeier: Partially fixed — the API-reference tags now read `contextual`, which is right. But the frozen **docs** version tags still print `typed-only`:

```
    2501  contextual   api-reference-7.6
     343  typed-only   docs-default-7.6.14
     264  typed-only   docs-default-6.28.11
```

A reader browsing the 7.6.14 guides reaches `docs-default-7.6.14` from that page's own contextual filter with nothing typed — exactly the case the new `contextual` set was introduced for. It only covers values of `apiReferenceTagsByVersionTag`, not the version tags that key it.

Cheap fix: add the non-served keys of that map to `contextual` alongside their values. Report-only, so entirely up to you whether it is worth the line.

.github/workflows/build-docs.yml
```diff
@@ -29,3 +29,18 @@ jobs:
...
+      # Both failure modes are silent in production.
+      - name: Test search facet filters
+        run: yarn test:search-facets
+
+      # Compares build/search-tags.json (what this build's widget will filter on)
+      # against the live Algolia index. Catches a docs-version rename quietly
+      # dropping content out of search - see scripts/verify-search-tags.cjs.
+      # Not --strict on PRs: a crawl still catching up after a deploy, a release
+      # build that renames the served version, or an Algolia outage all warn
+      # rather than blocking unrelated work.
```
moritzhartmeier: **Nothing in CI ever runs `--strict`, and there is no scheduled workflow — so both of the new downgrades are unconditional.**

`verify-search-tags.cjs` justifies its two escape hatches on `--strict` existing somewhere, and refers to "the schedule" three times:

- `releasePending` → "`--strict` fails either way so main and the schedule still see them"
- transport errors → "Under `--strict` (main, or the schedule, where someone is watching) it still fails"

Neither exists. This workflow runs the same non-strict `yarn verify:search-tags` on `pull_request` **and** on `push: main`, and no workflow in `.github/workflows/` has a `schedule:` trigger. `grep -r -- --strict .github/ package.json` matches only the comment above. So every condition the gate now downgrades is downgraded everywhere, permanently.

What that costs, using the incident this gate was built for: the release build WARNs (correctly). Push-to-main also runs pre-crawl, so it WARNs too. The unreachable API reference only becomes a FAIL on some later unrelated PR, once the crawl has caught up — and nobody reads WARN in a green log. The gate degrades from "catches it" to "might mention it eventually."

Adding `--strict` to the push-to-main run is *not* the fix — that run is also pre-crawl, so release merges would go red for exactly the reason `releasePending` exists. The fix is the daily `--strict` run the code already assumes:

```yaml
on:
  schedule:
    - cron: "0 6 * * *"
```

with a job that builds and runs `yarn verify:search-tags --strict`. Post-crawl, once a day, someone watching — which is precisely the contract the comments describe.

**Related, same theme:** the two new escape hatches overlap. If the file cap is hit (`INCONCLUSIVE`, no failure) while `releasePending` is true (both index checks → WARN), a genuine config drift passes fully green with no signal at all. Low probability at ~2,250 pages against a 4,000 cap, but the holes line up, and `--strict` is the only thing that would catch it.

moritzhartmeier (changes requested):
Re-verified `046d9720` by execution rather than by reading the commit message. **Both blockers are genuinely fixed**, and I have resolved 11 of the 12 threads.

**`rewriteVersionTag`** — my probe now returns two top-level entries with the duplicate gone. And the regression guard is real, not decorative: I ran the new test suite against the old `SearchBar` and it fails with `actual: 3, expected: 2`. The `isTagOrGroup` guard is a better fix than I suggested — it also blocks appending into a non-tag group, which I had not considered.

**The gate** — behaves exactly as claimed:

| run | result |
|---|---|
| current state | `OK`, exit 0 |
| 8.5.3 -> 8.5.4 | `WARN` x2, **exit 0** |
| 8.5.4 with `--strict` | `FAIL` x2, **exit 1** |

`missing` prints once now, and a typo'd `lastVersionTag` is still caught by the separate build-emission check, so `releasePending` does not open the hole I was worried about.

Also confirmed: shim gone and the gate now prints `1  UNREACHABLE  docs-default-current`; the factual correction landed in both the config SSOT block and the gate docstring with the cycle explanation; `tsc` clean; 7 facet tests pass; all 5 CI checks pass.

### One thing left, and I do think it should be fixed here

The `--strict` gap, inline. It is not a bug in what ships to readers, but it hollows out both of the fixes above: the release-PR fix and the transport-error fix are each justified by a strict run that does not exist anywhere. Fifteen lines of YAML and the contract the comments describe is actually true.

Left unresolved separately: the `typed-only` label, which is only partially fixed — report-only, your call.

Two figures for the record, since they moved: `api-reference-latest` is now **4,376** pages, not the 3,407 I measured on the 21st. The commit message has the current number; the code comment in `test-search-facets.cjs` still cites mine.

scripts/verify-search-tags.cjs
moritzhartmeier: Fixed. The `contextual` set now covers the version tags themselves, not just the API trees. Confirmed against the live index on a real build:

```
    2501  contextual   api-reference-7.6
     342  contextual   docs-default-7.6.14
     265  contextual   docs-default-6.28.11
```

Resolving.

.github/workflows/build-docs.yml
moritzhartmeier: Fixed, and I verified the semantics rather than just the wiring.

The daily `cron` plus `workflow_dispatch` plus `github.ref == 'refs/heads/main'` gives the post-crawl strict run the script is designed around, and a scheduled run's ref satisfies that expression, so there is no second code path to keep in step. Good.

I also checked the claim that `--strict` deliberately does *not* override `releasePending`, since that inverts what the previous head did. It holds, and both cycle events stay green either way:

| simulation | plain | `--strict` |
|---|---|---|
| patch 8.5.3 -> 8.5.4 | exit 0, 2 WARN | exit 0, 2 WARN |
| beta -> production | exit 0, 1 WARN | exit 0, 1 WARN |

And the suppression is backstopped, which was my remaining worry: pointing the manifest at a tag the build never emits still produces two hard FAILs (the `v8` routing check, and the build-emission check now scanning all 4,397 pages) even with `releasePending` active. So a genuinely broken config cannot hide behind it.

Resolving.

scripts/test-search-facets.cjs
```diff
@@ -0,0 +1,281 @@
...
+    name === "current"
+      ? path.join(__dirname, "..", "docs")
+      : path.join(__dirname, "..", "versioned_docs", `version-${name}`);
+  // Stated by the build. `docs-default-current` has no number in its name, so
+  // guessing one meant /next/ was never really checked.
+  const number = (manifest.versionNumberByTag || {})[versionTag];
+  assert.ok(number, `manifest states no version number for ${versionTag}`);
+  const line = number.split(".").slice(0, 2).join(".");
+  const own = `docs.scandit.com/${line}/data-capture-sdk`;
+  const unversioned = "docs.scandit.com/data-capture-sdk";
```
moritzhartmeier: Non-blocking, but the independence claim above is stronger than what the code does.

This needle is byte-identical to the config's (`docusaurus.config.ts:234`). The two scans differ in *traversal* — early-exit on one hit here, count both forms there — but they share the assumption that API links are written as absolute `docs.scandit.com/...` URLs.

So a root-relative link (`](/7.6/data-capture-sdk/...)`) is invisible to both, identically: that version silently maps to `api-reference-latest`, the crawler never produces a matching tag, and every test stays green. That is the exact bug class deriving-from-sources was introduced to prevent.

I checked — there are currently **0** root-relative API links in `docs/` or `versioned_docs/`, so the assumption holds today. Matching both forms (or asserting that the absolute form is the only one in use) would close it.

scripts/test-search-facets.cjs
```diff
@@ -0,0 +1,281 @@
...
+  check(`${versionTag} is mapped to the tree its own pages link to`, () => {
+    const forms = linkFormsIn(versionTag);
+    if (!forms) return; // version not checked out; nothing to contradict
+    const { ownHits, unversionedHits, line } = forms;
+    assert.ok(
+      ownHits === 0 || unversionedHits === 0,
+      `${versionTag} links to BOTH trees (${unversionedHits} unversioned, ` +
+        `${ownHits} to /${line}/). The config's scan early-exits on the first ` +
+        `own-line hit, which assumes these are mutually exclusive.`,
+    );
```
moritzhartmeier: Non-blocking, but worth knowing you have signed up for it: this couples docs *content* to the search gate.

A writer adding a single unversioned API link to any frozen version's page now fails `yarn test:search-facets` — and therefore the docs build — with a message about the config's scan early-exiting. That is a correct invariant, but it is a confusing thing to hit while editing a 7.6 page, and the person who hits it will not be the person who wrote this.

If the mutual exclusivity is genuinely required by the config's early-exit, the cheaper fix is to drop the early-exit and count both forms there too, so mixed links become a mapping decision rather than a build failure.

moritzhartmeier (approved):
## Approving

Verified on `2ba81930` with a real build rather than by reading the commit messages: build succeeds, **11/11** facet tests pass, the gate exits 0 against the live index, `tsc --noEmit` clean, all 5 CI checks green. Both blocking bugs from my first review are genuinely fixed, and I re-confirmed the `rewriteVersionTag` fix directly — including an adversarial case your new content-gate anticipates (language filter wrapped in an array; the API tag still lands in the tag group, not the language group).

Both release transitions stay green with and without `--strict`, and the `releasePending` suppression is backstopped: a manifest naming a tag the build never emits still produces two hard FAILs. Details in the two threads I just resolved.

### Two corrections to my own review

The four self-review passes found real defects, and three of them were in things I had verified and marked resolved. Worth recording so the next reader does not trust my earlier notes over the current code:

- **My file-cap headroom estimate was wrong.** I wrote "~2,250 pages, about two frozen versions of headroom" — I had counted source `.md` files. The build emits **4,397** `index.html`, already past the old 4,000 cap. So the `INCONCLUSIVE` path I asked for would have permanently masked the build-emission check rather than fixing it. Your fifth pass caught that; reading a 4 KB head per file instead of the whole thing is the right answer.
- **My `releasePending` verification was incomplete.** I only simulated the patch-during-beta transition and resolved the thread on that basis. Re-running the exact code I approved (`046d9720`) against a beta -> production manifest gives **exit 1 with a hard FAIL**, because `lastVersionTag` becomes `docs-default-current`, which the index already holds with one stale record. That is one of the three cycle events, and I should have tested it before resolving.
- Relatedly, I described the API mapping as "cycle-stable." It was correct only by coincidence — keyed off which version is *served* rather than what its pages *link to*. Deriving it from the sources is a real improvement over what I approved.

### Non-blocking notes

Two inline. The rest, none of which should hold this up:

- `yarn test:search-facets` now exits 1 without `build/search-tags.json`. Correct in CI, friction locally — a line in the README or a `pretest` hint would help.
- `docusaurus.config.ts:249` has `/.mdx?$/i` with an unescaped dot; the test has `/\.mdx?$/i`. Harmless in practice, in the file that derives search behaviour.
- `linkFormsIn` returning `null` makes those checks pass vacuously when a version is not on disk. Fine, just not a guarantee.
- The index counts move: `api-reference-latest` measured 3,407 -> 4,376 -> 3,908 across five days of crawling. Fixed `THIN`/`ORPHAN` thresholds against a moving number will produce occasional noise; if it does, that is why.

### One judgment call, yours not mine

The verification layer is now considerably larger and more intricate than the fix it guards — `verify-search-tags.cjs` grew ~423 lines across four passes, and three of those passes were fixing checks that reported OK on the very states they were written to catch. That is a good signal about the diligence and a slightly uncomfortable one about the complexity: a gate against silent failure is itself a place where silent failure hides, and this one has now had more iterations than the bug it protects against.

Not a reason to block. But if you want the gate reviewed on its own terms rather than as an appendix to a search fix, splitting it into its own PR before merge would be the moment. Your call — approving either way.

PR #429, opened 2026-08-20
fix(docs): clear the Vale findings, and fix the rule that produced most of them

eugenia-scandit:
233 findings across 57 files. Only 79 were real.

Google.Spacing matches `[a-z][.?!][A-Z]` anywhere, which in API documentation is overwhelmingly a code identifier rather than a sentence: SymbologySettings.Checksums, context.SetFrameSourceAsync(), Scandit.DataCapture.Core.Maui. Of its 155 findings, 154 were identifiers and exactly one was a real missing space. Acting on the other 154 would have corrupted a symbol name each time - so the rule needed fixing, not the prose.

styles/Scandit/Spacing.yml replaces it (Google.Spacing = NO in .vale.ini) and expresses the sentence boundary positionally, since RE2 has no lookaround: a whole lowercase word, the punctuation, then a single capitalised word, with whitespace on both outer edges. An identifier fails at least one part. It still catches the one real case ("smart device.It requires").

An earlier attempt used TokenIgnores instead. It silently swallowed "e.g." and "i.e." too, disabling Google.Latin - 46 real findings vanished and looked like success. Worth recording: the cheap fix here quietly removes a rule you wanted.

The remaining 79 are fixed as prose:
- 46 Latin abbreviations (e.g. / i.e. spelled out)
- 18 banned marketing words - "seamless user experience" -> "responsiveness", "ensures seamless integration into your workflow" -> "fits into your workflow", "simply change" -> "change", and so on: what the sentence claims, minus the adjective that claimed it
-  8 optional plurals: item(s) -> items, plugin(s) -> plugins
-  2 ordinals (3rd-Party -> Third-Party), 2 em dashes, 1 quote placement, 1 unit spacing, 1 missing space after a full stop

Replacements are applied outside fenced and inline code only, so no snippet changed meaning; the one finding inside a code comment was reworded by hand. Vale now reports 0 findings across all 117 files, and the site builds.


styles/Scandit/Spacing.yml
```diff
@@ -0,0 +1,115 @@
...
+# on this branch 0 are, because it fixes all four - in docs/ and in the versioned
+# trees that actually publish them. These numbers describe commits, not a
+# backlog, so re-measure rather than expecting the rule to fire today. Two of
+# those 3 were in docs/partials/, which the gate did not lint at the time, so the
+# rule's own evidence was unenforceable. That hole is closed in this branch -
+# though only the _symbology-properties.mdx one was ever on a published page, the
+# other partial being unreferenced.
+tokens:
+  - '[a-z][.?!] {2,}[A-Z]'
+  - '(?:^|\s)[a-z]+[.?!][A-Z][a-z]+(?:\s|$)'
```
moritzhartmeier: This rule flags any un-backticked prose reference like `settings.Checksums` as a missing-space violation, and since `level: error` + `MinAlertLevel = error` makes this a blocking check, the generic message ("'%s' should have one space") leads an author straight to a corrupting fix: inserting a space turns `settings.Checksums` into `settings. Checksums`.

Example: "Configure settings.Checksums for this feature." (identifier not backticked) triggers the rule and suggests corrupting the identifier. The rule's own comment notes current corpus exposure is 0, but that's only because docs happen to backtick identifiers by convention — nothing here checks for or requires that.

moritzhartmeier (commented):
Automated review: 4 findings (2 correctness, 1 stale comment, 1 maintainability nit).

styles/Scandit/Spacing.yml
moritzhartmeier: Token 2 (`(?:^|\s)[a-z]+[.?!][A-Z][a-z]+(?:\s|$)`) consumes the trailing whitespace delimiter, so on a line with two separate missing-space violations only the first is reported — the second is silently missed.

Verified with `vale` against "Two in a row: alpha.Beta gamma.Delta end." — only `alpha.Beta` is reported; `gamma.Delta` is not. An author who fixes the one reported issue and re-pushes can merge a page that still has the second, now-invisible spacing bug.

.github/workflows/docs-gate.yml
```diff
@@ -2,8 +2,30 @@ name: Docs Quality Gate
...
+# page's entire backlog. That is deliberate - it is what stops the debt growing
+# - but it is the gate's most surprising behaviour, so budget for it when
+# touching an old page.
+#
+# Two things are outside the gate's reach, by construction rather than by
+# accident:
+#
+#   - versioned_docs/. The gate diffs `-- docs` only, so the 1,585 files under
+#     versioned_docs/ are never linted - including the tree that
+#     docusaurus.config.ts serves at the site root via lastVersion. A defect
```
moritzhartmeier: This comment says versioned_docs/ "is the tree that docusaurus.config.ts serves at the site root via lastVersion," but that was only true when this PR's earlier commit was written (lastVersion pinned to 8.5.3). The later merge commit on this branch (e34b52acc, "Merge origin/main: the 8.6 release retires the 8.5.3 snapshot") explicitly says the 8.6 release deleted versioned_docs/version-8.5.3 and set `DOCS_LAST_VERSION = "current"`, so docs/ is now what's served at root — confirmed in docusaurus.config.ts on this branch's head.

This comment wasn't updated after that merge, so it now contradicts both the merge commit's own message and the live config. Worth rephrasing to something version-agnostic, e.g. "never linted — this matters whenever lastVersion points at a snapshot rather than current; today it doesn't, so docs/ fixes are live at the root immediately, but that flips at the next beta cycle," rather than asserting unconditionally that versioned_docs/ is the root-served tree.

styles/Scandit/Spacing.yml
```diff
@@ -0,0 +1,115 @@
+extends: existence
+message: "'%s' should have one space."
+link: 'https://developers.google.com/style/sentence-spacing'
+level: error
+nonword: true
+# Replaces Google.Spacing, which is disabled in .vale.ini.
```
moritzhartmeier: Minor: this rule carries ~110 lines of commentary citing exact alert counts and false-positive examples measured at specific past commits, with no mechanism keeping them current (the comment itself says to re-measure rather than trust them). A future editor changing an ignore pattern or adding/removing a page will silently invalidate these numbers, and a reader skimming the file may take them as current guidance.

styles/Scandit/Spacing.yml
moritzhartmeier: Fixed in caef1b82a: the message now reads "needs one space after the sentence punctuation - unless it is a code identifier, which belongs in backticks," so an author hitting this on an un-backticked identifier is pointed at the backtick fix instead of a corrupting space-insert. Verified with vale against "Configure settings.Checksums for this."

moritzhartmeier: Fixed in caef1b82a: token 2's trailing (?:\\s|$) became a lookahead (?=\\s|$), so it no longer consumes the delimiter the next match needs. Verified with vale against "Two in a row: alpha.Beta gamma.Delta end." — both alpha.Beta and gamma.Delta are now reported.

.github/workflows/docs-gate.yml
moritzhartmeier: Fixed in 3361425510: the comment is now version-agnostic — it says to read docusaurus.config.ts rather than assume, and correctly describes both states (today lastVersion = "current" so docs/ is live at root; during a beta cycle it'd be the pinned snapshot).

styles/Scandit/Spacing.yml
moritzhartmeier: Fixed in 3361425510 (and caef1b82a for Spacing.yml): hardcoded counts replaced with "run vale docs for the count, don't trust a figure quoted here," and the long measurement narrative was trimmed to describe mechanism rather than numbers that go stale.

moritzhartmeier (approved):
All 4 findings from the initial review confirmed fixed and verified (message text and vale output checked directly, not just commit messages). Approving.

PR #430, opened 2026-08-20
fix(docs): validate the singular framework field and collapse five framework maps into one

eugenia-scandit:
Closes the ticket on the unvalidated `framework` field.

## The problem

`docs-schema.yml` defined the plural `frameworks` array with an enum but did not
define the singular `framework` field at all. The schema allows undeclared
fields, so it got no validation and its values drifted. Components that read it
silently found no match and fell back to their default — always iOS. No crash,
no error.

## What this changes

- The enum in `docs-schema.yml` is the single source of truth; singular and
  plural share one vocabulary.
- 117 pages normalized: 92 spelling drifts (`netIos` / `netAndroid` / `react`),
  and 25 `express` / `bolt` pages which were never frameworks — they live under
  `docs/hosted/`, so they are now `framework: hosted` with the value they carried
  moved to `product:`.
- Five copies of the framework list replaced by one registry
  (`src/constants/frameworks.ts`), which `UNRELEASED_FRAMEWORK_SLUGS` also
  derives from.

## Three things the ticket could not have known

- **`kmp` — 37 pages**, more than any row in the ticket's table. Not drift: it is
  the published `/sdks/kmp/` tree, it is first-class in the code, and
  "unreleased" is already tracked by a separate mechanism. The enum was stale, so
  `kmp` was added to it rather than the pages being rewritten.
- **`FW_TO_PJ` does not exist** — PR #417 is unmerged. There were five copies,
  not three.
- **A live bug:** `FRAMEWORK_MAPPING` had no `linux` entry, so
  `parseSdksRoute()` resolved no framework at all for every `/sdks/linux/` page.

## Why nobody ever saw an error

`docs-gate` is a ratchet over changed files. Those 37 pages had been violating
the plural enum for months and were never once reported. So
`yarn verify:frameworks` (in CI) runs over the whole corpus and over the
registry: a value outside the enum is an error, and the registry and schema are
compared in both directions.

Plus `yarn test:frameworks` — 9 assertions pinning the membership of every
derived map, because the risk in this refactor is not a crash but a map quietly
gaining or losing an entry.

## Separately

`docs-gate` now skips prose checks for files whose body did not change. The
mechanical frontmatter pass across 117 files dragged in 233 pre-existing Vale
findings (measured: 233 before, 233 after). Any file with a body edit is still
checked in full — verified by adding one sentence and watching both new findings
appear while the other 116 stayed skipped.

**The ticket's test:** a page set to `framework: netIos` fails the gate before
normalization and passes after — verified in both directions.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -27,5 +27,12 @@ jobs:
       - name: Install dependencies
         run: yarn install --frozen-lockfile
 
+      # Whole-corpus check: the docs gate is a ratchet on changed files, so it
+      # cannot see a pre-existing violation nobody touched. This is what makes a
+      # framework value outside the enum - in docs OR in the code maps that read
+      # it - break the build instead of silently mis-resolving.
+      - name: Verify framework identifiers
+        run: yarn verify:frameworks
+
       - name: Build documentation
         run: yarn build
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -34,5 +34,10 @@ jobs:
       - name: Verify framework identifiers
         run: yarn verify:frameworks
 
+      # Pins the membership of every derived framework map. The refactor risk
+      # here is not a crash, it is a map quietly gaining or losing an entry.
+      - name: Test framework registry
+        run: yarn test:frameworks
+
       - name: Build documentation
         run: yarn build
--- a/scripts/docs-gate/index.cjs
+++ b/scripts/docs-gate/index.cjs
@@ -10,6 +10,8 @@ const { loadSchema, validateFile } = require("./frontmatter.cjs");
 const { checkLinks } = require("./links.cjs");
 
 const ROOT = process.cwd();
+// The ratchet base changedDocs() resolved, reused by frontmatterOnly().
+let lastRatchetBase = "";
 
 function sh(cmd) {
   // Surface git's stderr (do not swallow it) so a failed ratchet lookup is
@@ -45,10 +47,53 @@ function changedDocs() {
   try { out += "\n" + sh("git diff --name-only --diff-filter=ACMR -- docs"); } catch {}
   try { out += "\n" + sh("git diff --cached --name-only --diff-filter=ACMR -- docs"); } catch {}
   try { out += "\n" + sh("git ls-files --others --exclude-standard -- docs"); } catch {}
+  lastRatchetBase = base; // reused by frontmatterOnly()
   const files = [...new Set(out.split(/\r?\n/).filter(Boolean))];
   return files.filter(
     (f) => /\.(md|mdx)$/i.test(f) && !path.basename(f).startsWith("_") && fs.existsSync(path.join(ROOT, f))
   );
+}
+
+/** Everything after the frontmatter block, or the whole file if there is none. */
+function bodyOf(text) {
+  // git show hands back the repo blob with LF while the Windows working copy
+  // has CRLF; without this every file compares as changed and the skip never fires.
+  text = text.replace(/\r\n/g, "\n");
+  if (!text.startsWith("---")) return text;
+  const end = text.indexOf("\n---", 3);
+  return end === -1 ? text : text.slice(end + 4);
+}
+
+/**
+ * Files whose diff against the ratchet base touches frontmatter only.
+ *
+ * The ratchet checks a whole file as soon as a PR touches one line of it, which
+ * is right for prose someone is actually editing and wrong for a mechanical
+ * metadata pass: normalizing a `framework:` value across 117 pages dragged in
+ * 233 pre-existing Vale findings that the change neither caused nor altered.
+ * Nobody writes prose in frontmatter, so when the body is byte-identical to the
+ * base there is no prose to review, and the prose checks are skipped for that
+ * file. Schema and link checks still run on every changed file.
+ */
+function frontmatterOnly(files, base) {
+  const out = new Set();
+  if (!base) return out;
+  for (const f of files) {
+    let before;
+    try {
+      before = sh(`git show ${base}:${f}`);
+    } catch {
+      continue; // new file - it is all new, check everything
+    }
+    let after;
+    try {
+      after = fs.readFileSync(path.join(ROOT, f), "utf8");
+    } catch {
+      continue;
+    }
+    if (bodyOf(before).trim() === bodyOf(after).trim()) out.add(f);
+  }
+  return out;
 }
 
 function findVale() {
@@ -111,17 +156,27 @@ function main() {
   if (files.length === 0) { console.log("docs-gate: no changed docs — nothing to check."); process.exit(0); }
   console.log(`docs-gate: checking ${files.length} changed doc(s)…\n`);
 
+  // Prose checks only look at files whose body actually changed.
+  const metaOnly = frontmatterOnly(files, lastRatchetBase);
+  const proseFiles = files.filter((f) => !metaOnly.has(f));
+  if (metaOnly.size) {
+    console.log(
+      `docs-gate: ${metaOnly.size} file(s) changed frontmatter only - ` +
+        `skipping prose checks for them (body identical to base).\n`,
+    );
+  }
+
   const schema = loadSchema(path.join(ROOT, "docs-schema.yml"));
   let findings = [];
   for (const f of files) {
     findings.push(...validateFile(path.join(ROOT, f), schema).map((x) => ({ ...x, file: f })));
     findings.push(...checkLinks(path.join(ROOT, f)).map((x) => ({ ...x, file: f })));
   }
-  findings.push(...runCspell(files));
+  if (proseFiles.length) findings.push(...runCspell(proseFiles));
 
   const vale = findVale();
   if (vale) {
-    findings.push(...runVale(files, vale));
+    if (proseFiles.length) findings.push(...runVale(proseFiles, vale));
   } else if (process.env.CI) {
     // In CI, a missing Vale must fail — otherwise the headline prose-style check
     // silently no-ops while the job stays green. Locally it's still advisory.
--- a/src/components/utils/frameworks.ts
+++ b/src/components/utils/frameworks.ts
@@ -3,6 +3,7 @@ import {
   FRAMEWORKS,
   ROUTED_FRAMEWORKS,
   AGENT_SKILL_FRAMEWORKS,
+  frameworkFromRouteTail,
 } from '@site/src/constants/frameworks';
 
 // localStorage key the homepage framework selector writes and the shared
@@ -45,23 +46,30 @@ export interface SdksRouteInfo {
 }
 
 export function parseSdksRoute(pathname: string): SdksRouteInfo {
-  // Tolerate a leading docs-version segment (/next/, /7.6.14/, ...): only the
-  // version served at the site root has none, so matching on /sdks/ alone would
-  // silently fail on every other version.
-  const match = pathname.match(
-    /^(?:\/(?:next|\d+\.\d+\.\d+))?\/sdks\/((?:net\/)?[^\/]+)\/([^\/]+)(?:\/([^\/]+))?/,
-  );
+  // Anchored, with an optional docs-version segment (/next/, /7.6.14/, ...):
+  // only the version served at the site root has none. Left unanchored,
+  // /foo/sdks/ios/... would parse as a real product route.
+  const match = /^(?:\/(?:next|\d+\.\d+\.\d+))?\/sdks\/(.+)$/.exec(pathname);
   if (!match) return {};
 
-  const rawFramework = match[1];
-  const rawProduct = match[2];
-  const last = match[3];
+  // Which framework the tail belongs to is the registry's business - it owns
+  // `routeSegment`, including the two-segment .NET routes this function used to
+  // hardcode as `(?:net\/)?` and then undo with `.replace('/', '-')`. That copy
+  // is why FeatureList's own regex could disagree with this one.
+  const def = frameworkFromRouteTail(match[1]);
+  if (!def || def.routeSegment === null) return {};
 
-  const frameworkSlug = rawFramework.replace('/', '-');
-  const framework = FRAMEWORK_MAPPING[frameworkSlug];
+  const rest = match[1].slice(def.routeSegment.length).replace(/^\//, '');
+  const [rawProduct, last] = rest.split('/');
+  // A product segment is required: /sdks/ios/ on its own is not a product page.
+  if (!rawProduct) return {};
+
   const product = URL_PRODUCT_MAPPING[rawProduct] || rawProduct;
-
-  return { framework, product, lastSegment: last };
+  // `lastSegment` is omitted rather than set to undefined, so the returned shape
+  // matches what the previous regex produced for a two-segment route.
+  return last
+    ? { framework: def.display, product, lastSegment: last }
+    : { framework: def.display, product };
 }
 
 // Maps the ?framework= query slug used on the homepage to an agent-skills URL path.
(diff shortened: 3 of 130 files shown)


.github/workflows/build-docs.yml
```diff
@@ -45,6 +45,25 @@ jobs:
...
+      # here is not a crash, it is a map quietly gaining or losing an entry.
+      - name: Test framework registry
+        run: yarn test:frameworks
+
+      # The gate scripts' own readers and, through a fixture tree, the whole of
+      # verify-frameworks. Every defect found across nine review rounds on this
+      # change lived in one or the other, and each made a gate print OK while a
+      # check went unperformed.
+      - name: Test docs-gate readers
+        run: yarn test:docs-gate
```
moritzhartmeier: **Blocking.** This is the only workflow that runs `test:docs-gate`, and it has no Vale install step — only `docs-gate.yml` installs Vale. `ubuntu-latest` has none preinstalled, and nothing in `yarn install` provides it (`scripts/setup-vale.cjs` only checks). Meanwhile `scripts/test-docs-gate.cjs` hard-asserts it:

```js
assert.ok(
  valePresent,
  "vale is not on PATH; these rows assert prose alerts and would pass vacuously without it",
);
```

Reproduced by running the suite with Vale removed from `PATH`: `1 failed, 36 passed`, exit 1. So Build Documentation fails on every PR, every push to `main`, and the daily schedule.

Either add the same pinned Vale install step `docs-gate.yml` uses, or move `test:docs-gate` into `docs-gate.yml` where Vale is already present.

scripts/test-docs-gate.cjs
```diff
@@ -0,0 +1,1221 @@
...
+  const { build } = require("./fixtures/docs-gate-fixture.cjs");
+  const script = path.join(ROOT, "scripts", "docs-gate", "index.cjs");
+  withTempDir((dir) => {
+    const repo = path.join(dir, "repo");
+    for (const [mutation, expected] of DOCS_GATE_CASES) {
+      build(repo, mutation);
+      let code = 0;
+      let out = "";
+      try {
+        out = execFileSync(process.execPath, [script], {
```
moritzhartmeier: **Blocking.** No `env` is passed here, so the child inherits GitHub's environment — including `GITHUB_BASE_REF`.

`build-docs.yml` triggers on `pull_request: branches: [main, 'release/**']`. On a release-branch PR, `GITHUB_BASE_REF=release/8.6`, so `changedDocs()` resolves `baseBranch = origin/release/8.6` — a ref that does not exist inside the self-contained fixture repo. With `CI=true` the gate takes its fail-loud path and `process.exit(2)`, and the `(clean)` row (which expects exit 0) fails.

Reproduced:

```
$ CI=true GITHUB_BASE_REF=release/8.6 node scripts/test-docs-gate.cjs
(clean): expected a clean pass, got 2
cannot resolve ratchet base 'origin/release/8.6'
```

It only passes today by luck, because the default `origin/main` happens to be the ref the fixture creates. Suggest passing an explicit `env` with `GITHUB_BASE_REF` (and ideally `CI`) deleted, the way the `verify-frameworks` rows already scope theirs.

src/components/utils/frameworks.ts
```diff
@@ -48,47 +46,60 @@ export interface SdksRouteInfo {
...
+  // Which framework the tail belongs to is the registry's business - it owns
+  // `routeSegment`, including the two-segment .NET routes this function used to
+  // hardcode as `(?:net\/)?` and then undo with `.replace('/', '-')`. That copy
+  // is why FeatureList's own regex could disagree with this one.
+  const def = frameworkFromRouteTail(match[1]);
+  // The null check is unreachable today - frameworkFromRouteTail searches only
+  // the routed frameworks, so `hosted` cannot come back from it. Kept because
+  // the next line dereferences routeSegment: if the routed set ever widens,
+  // the alternative to this branch is a runtime throw, not a type error.
+  if (!def || def.routeSegment === null) return {};
```
moritzhartmeier: **Non-blocking, but worth a look.** Returning `{}` here now drops `product` as well as `framework`. Previously an unrecognized framework segment still yielded `{framework: undefined, product, lastSegment}`.

In `src/theme/DocItem/Content/index.tsx` that flips `isKnownProductPage` from true to false, so instead of the product callout (which rendered nothing, since `resolvedFramework` was undefined) the page falls through to `<SkillsCallout variant="shared" />`, which defaults to iOS.

Concretely: add a new SDK tree under `docs/sdks/<new>/<known-product>/` before adding its registry entry, and every page in it gains an iOS-pointing Agent Skills banner where previously none appeared. No tree in the repo hits this today (the versioned `xamarin` snapshots already produced no known product), so this is a latent behaviour change rather than a live defect — but the failure mode is silent, which is exactly what the rest of this PR is trying to eliminate.

moritzhartmeier (changes requested):
Reviewed the full diff (136 files) against `main`, and verified behaviour in a throwaway worktree: `tsc --noEmit` clean, full `docusaurus build` green, `yarn verify:frameworks` / `test:frameworks` / `test:docs-gate` green, and I hand-checked the new docs-gate frontmatter cap both ways (a banned word in a `description` is still reported; a banned word in a body still triggers a full-file lint).

The refactor itself checks out. `FRAMEWORK_MAPPING`, `QUERY_FRAMEWORK_TO_PATH`, `HOMEPAGE_FRAMEWORK_ALIASES`, `FRAMEWORK_URL_PATH` and `FRAMEWORK_SLUG` all derive to exactly their previous membership plus the intended `linux` addition, and all 117 frontmatter edits agree with their directory tree.

The two blocking problems are in CI wiring rather than site code, and both reproduce locally:

1. **`build-docs.yml` runs `test:docs-gate` without installing Vale** — the suite hard-asserts Vale is on `PATH`, so this workflow goes red on every PR, every push to `main`, and the daily schedule.
2. **The docs-gate fixture inherits `GITHUB_BASE_REF`** — so the suite fails on any PR targeting `release/**`, which this same workflow triggers on.

There is also one non-blocking latent behaviour change in `parseSdksRoute`, noted inline. Details in the three inline comments.

.github/workflows/build-docs.yml
eugenia-scandit: Confirmed and fixed in 4d796f8b7 — moved into `docs-gate.yml` rather than duplicating the installer.

Taking the second of your two options. The deciding detail is that the suite needs the Vale **binary** only, not the styles: `scripts/fixtures/docs-gate-fixture.cjs` writes its own `.vale.ini` with no `Packages` line, precisely so the rows never hit the network. So it has no dependency on `docs:gate:setup` and can sit anywhere in that job.

Duplicating the pinned install block would have put `VALE_VERSION=3.15.1` in two workflows with nothing tying them together, and the whole point of pinning it is that the rules and the frontmatter cap were measured against that version. One copy, one job.

It runs before `yarn docs:gate` deliberately: if the gate's own readers are broken, that is worth knowing before reading anything the gate reports.

What moves: the suite no longer runs on push-to-main, schedule or workflow_dispatch. That is fine — these are unit tests over scripts, and the scripts only change through PRs, which `docs-gate.yml` covers on both `main` and `release/**`. `verify:frameworks` and `test:frameworks` need no Vale and stay in `build-docs.yml`, so the whole-corpus check still runs on every trigger.

scripts/test-docs-gate.cjs
eugenia-scandit: Confirmed and fixed in 4d796f8b7.

Reproduced your exact case before touching it:

```
$ CI=true GITHUB_BASE_REF=release/8.6 node scripts/test-docs-gate.cjs
(clean): expected a clean pass, got 2
cannot resolve ratchet base 'origin/release/8.6'
1 failed, 36 passed
```

And your reading of why it passes today is right — `scripts/fixtures/docs-gate-fixture.cjs:104` creates exactly one ref, `refs/remotes/origin/main`, so the default is the only value that resolves.

The child env now clears `GITHUB_BASE_REF`. After: 37 passed, both with `CI=true GITHUB_BASE_REF=release/8.6` and with a bare environment.

One deviation from your suggestion, and I'd rather flag it than let it pass unnoticed: you suggested deleting `CI` as well, and I've **pinned** it to `1` instead. Deleting it leaves the rows testing a different code path locally than in the job — with `CI` unset, an unresolvable base falls through to `base = ""`, `git diff` collapses to a working-tree diff, and in a clean checkout that is empty and reads as a pass. That is the vacuous-pass failure mode this suite exists to catch, so it should be loud in both places. Pinning it also makes the rows deterministic rather than dependent on ambient environment either way.

Nothing else in the gate branches on `CI` — only `index.cjs:44` and `index.cjs:406`, and the latter is the missing-Vale path, which the suite's own Vale assertion already rules out. Happy to switch to a plain delete if you'd rather keep it minimal.

src/components/utils/frameworks.ts
eugenia-scandit: Confirmed, and fixed in 4d796f8b7 rather than left as a latent change.

Your trace is exact. I checked the old implementation on `main` to be sure of what the behaviour was:

```js
const framework = FRAMEWORK_MAPPING[frameworkSlug];   // undefined for an unknown segment
const product = URL_PRODUCT_MAPPING[rawProduct] || rawProduct;
return { framework, product, lastSegment: last };     // product survived
```

So an unrecognised framework yielded `{framework: undefined, product, lastSegment}`, `isKnownProductPage` stayed true, and `<SkillsCallout variant="product" />` rendered nothing because it bails on a missing `resolvedFramework`. Returning `{}` turns "no banner" into "the wrong framework's banner" — iOS, on every page of the new tree.

`parseSdksRoute` now resolves the framework to `undefined` and keeps the product. One segment is the right assumption in that fallback: every multi-segment `routeSegment` is a registry entry by construction, so a tail that matched none of them cannot be one.

The part I'd have missed without your comment is that `PARSE_SDKS_ROUTE_BASELINE` had no row for this shape at all — it pins every *registered* framework and the empty cases, so the regression had nowhere to show up. Three rows now cover it:

```js
["/sdks/newsdk/barcode-capture/intro", { product: "barcode-capture", lastSegment: "intro" }],
["/sdks/newsdk/barcode-capture",       { product: "barcode-capture" }],
["/sdks/newsdk/",                      {}],
```

Mutation-tested rather than assumed: restoring `if (!def || def.routeSegment === null) return {}` fails them.

`framework` is omitted rather than set to `undefined`, matching how `lastSegment` is already handled, so `deepStrictEqual` keeps agreeing with the rest of the table.

.github/workflows/build-docs.yml
moritzhartmeier: Fixed — the step moved to `docs-gate.yml`, where the pinned Vale 3.15.1 install runs ahead of it. I checked the resulting step order, and confirmed the hard assert is still in place (Vale off `PATH` still gives `1 failed, 36 passed`), so this is the move rather than a weakened assert. Resolving.

scripts/test-docs-gate.cjs
moritzhartmeier: Fixed. Reproduced both sides: at `16a6093ca`, `CI=true GITHUB_BASE_REF=release/8.6 node scripts/test-docs-gate.cjs` gave `1 failed, 36 passed`; at head it gives `37 passed`. Pinning `CI: "1"` in `gateEnv` alongside the `GITHUB_BASE_REF` delete is the right call — it keeps the loud exit-2 path instead of letting an unresolvable base degrade to a vacuously clean working-tree diff. Resolving.

src/components/utils/frameworks.ts
moritzhartmeier: Fixed. Probed the real module: `/sdks/newsdk/barcode-capture/intro` now returns `{product, lastSegment}` with the framework absent, and `SkillsCallout/index.tsx:273` returns null in that state, so the iOS-defaulting shared callout is genuinely avoided. Registered routes are unchanged. Resolving.

scripts/docs-gate/index.cjs
```diff
@@ -17,6 +17,18 @@ function sh(cmd) {
...
 
+/**
+ * Like sh(), but for a lookup whose failure is EXPECTED and meaningless.
+ *
+ * `git show <base>:<path>` cannot succeed for a file the PR adds, and sh()
+ * inherits git's stderr on purpose, so a PR adding forty pages printed forty
+ * `fatal: path ... exists on disk, but not in ...` lines before the gate said
+ * anything - which reads as a crashed gate rather than as forty new files.
+ */
+function shQuiet(cmd) {
```
moritzhartmeier: **Non-blocking follow-up.** `shQuiet()` is now dead code. `c22e1376e` replaced its only call site — the `git show <base>:<path>` lookup in `frontmatterOnly` — with `execFileSync`, so the function is defined with zero callers:

```
$ git grep -n shQuiet
scripts/docs-gate/index.cjs:28:function shQuiet(cmd) {
scripts/docs-gate/index.cjs:160:      // shQuiet: a file the PR ADDS has no base blob, and that failure is
scripts/docs-gate/index.cjs:161:      // both expected and meaningless here - see the note on shQuiet.
```

The two remaining mentions are its own docblock and the comment at 160-161, which still tells the reader to "see the note on shQuiet" for a helper no longer in the path. Nothing misbehaves at runtime — the `execFileSync` call carries the same `stdio: ["ignore", "pipe", "ignore"]`, so stderr stays quiet for the added-file case exactly as before.

Suggest deleting the function and rewording 160-161 to describe the `stdio` on the call itself. It's the same orphaned-state cleanup `cd4574004` did for `lastRatchetBase`.

moritzhartmeier (commented):
The three findings from my earlier review are all fixed — I reproduced each original failure condition against the new code rather than reading the diff, and replied on each thread with what I checked before resolving it. Everything is green at head: `test:docs-gate` 37, `test:frameworks` 16, `verify-frameworks` OK, `tsc --noEmit` clean.

One non-blocking leftover from `c22e1376e`, inline. Also worth a note, though I don't think it needs action: `test:docs-gate` now lives only in `docs-gate.yml`, which is `pull_request`-only, so the gate's own tests no longer run on push-to-`main` or the daily schedule that `build-docs.yml` carries. PR coverage is identical (both workflows target `main` and `release/**`) and main-only breakage would surface on the next PR, so it reads as an acceptable trade rather than a regression — flagging it only so the narrowing is a deliberate choice.

moritzhartmeier (approved):
Approving. Both blockers are fixed and verified by reproducing the original failure conditions, and the four cleanup commits check out — the deletions (`FRAMEWORK_BY_SLUG`/`FRAMEWORK_BY_DISPLAY`, `lastRatchetBase`) have no surviving references, and `tsc --noEmit` plus every suite is green at head.

The `shQuiet` cleanup above is a nit, not a merge blocker — happy for it to land here or in a follow-up.

scripts/docs-gate/index.cjs
eugenia-scandit: Correct on all counts, and fixed — but in **#444** rather than here, deliberately.

I had pushed the fix onto this branch, then realised it dismissed your approval: you approved `cd4574004`, and any commit after that resets `reviewDecision` to `REVIEW_REQUIRED`. Since you called this a non-blocking follow-up, trading a re-review for it seemed like the wrong deal. So this branch is rewound to exactly the tree you approved, and the fix lives in #444, based on this branch so its diff is just the two files. GitHub retargets it to `main` once this merges.

#444 does what you describe: deletes `shQuiet`, and rewrites the comment at 160 so it names what actually suppresses the noise — `stdio: ["ignore", "pipe", "ignore"]` — and why there is noise: a file the PR *adds* has no base blob, so git prints one `fatal:` line per new page. Your framing of the cost is the part I kept, because it is the real one: the danger is not the dead function, it is the comment telling the next reader to go looking for it, and someone "restoring" the call reintroduces the shell interpolation `c22e1376e` removed on purpose.

Resolving this so it stops blocking the merge — the work is tracked in #444, not dropped. Reopen if you would rather it land here and take the re-review.

PR #431, opened 2026-08-20
ci: gate the API reference against duplicate-content across versions

eugenia-scandit:
The API reference is published once per major.minor line, so the same symbol page exists at /6.28/data-capture-sdk/X, /7.6/data-capture-sdk/X and the unversioned /data-capture-sdk/X. None of them declares which is canonical, so Google treats them as independent pages and picks whichever it likes - usually the oldest, because it carries the most history and inbound links. That is how /6.28/.../aamva-barcode-result.html came to outrank current docs.

The rule this checks, per versioned page:

  unversioned counterpart returns 200  ->  canonical pointing at it
  unversioned counterpart returns 404  ->  robots noindex

The second branch is the one easy to get wrong. An API removed since that line has no current equivalent, so a canonical would point at a 404 and Google would ignore it; that page needs de-indexing, not redirecting. Both cases occur in practice - the first sampled run found 4 of one and 3 of the other.

Nothing here is version-specific, which is the point: the canonical target is always the unversioned URL, which is by definition the current line. A release never changes what this expects and there is no version constant to maintain.

URLs come from the versioned API links the built site already contains (2,049 distinct across /6.28/ and /7.6/), sampled deterministically so CI checks the same pages each run rather than drifting.

Warns rather than fails. The API-reference HTML is generated outside this repository, so this gate can only observe it - it is merged now so the finding is visible and tracked, and takes --strict once the generator emits the tags.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -29,3 +29,11 @@ jobs:
 
       - name: Build documentation
         run: yarn build
+
+      # The versioned API reference publishes the same page under several URLs
+      # and declares no canonical form, so Google picks one - often the oldest.
+      # Warns for now: the API-reference HTML is generated outside this repo, so
+      # this cannot pass until that generator emits canonical/noindex. Add
+      # --strict to make it blocking once it does.
+      - name: Verify API-reference canonical tags
+        run: yarn verify:api-reference-seo
--- a/scripts/verify-api-reference-seo.cjs
+++ b/scripts/verify-api-reference-seo.cjs
@@ -1,0 +1,196 @@
+#!/usr/bin/env node
+"use strict";
+/**
+ * Duplicate-content gate for the versioned API reference.
+ *
+ * The API reference is published once per major.minor line, so the same symbol
+ * page exists at /6.28/data-capture-sdk/X, /7.6/data-capture-sdk/X and the
+ * unversioned /data-capture-sdk/X. None of those pages declares which one is
+ * canonical, so Google treats them as independent and picks whichever it likes -
+ * usually the oldest, because it has the most history and inbound links. That is
+ * how /6.28/.../aamva-barcode-result.html ended up outranking current docs.
+ *
+ * The rule this checks, per versioned page:
+ *
+ *   unversioned counterpart returns 200  ->  <link rel="canonical"> pointing at it
+ *   unversioned counterpart returns 404  ->  <meta name="robots" content="noindex">
+ *
+ * The second branch matters: an API removed since that line has no current
+ * equivalent, so a canonical would point at a 404 and Google would ignore it.
+ * That page needs de-indexing, not redirecting.
+ *
+ * Nothing here is version-specific. The canonical target is always the
+ * unversioned URL, which is by definition the current line, so a release never
+ * changes what this expects and there is no version constant to maintain.
+ *
+ * The API-reference HTML is generated outside this repository, so this gate can
+ * only observe it. It warns by default and fails only with --strict, so it can
+ * be merged before the generator is fixed and switched to blocking after.
+ *
+ * Usage: node scripts/verify-api-reference-seo.cjs [--strict] [--sample N]
+ */
+
+const fs = require("fs");
+const path = require("path");
+
+const ROOT = path.join(__dirname, "..");
+const BUILD = path.join(ROOT, "build");
+const ORIGIN = "https://docs.scandit.com";
+
+const argv = process.argv.slice(2);
+const strict = argv.includes("--strict");
+const sampleSize = argv.includes("--sample")
+  ? Number(argv[argv.indexOf("--sample") + 1])
+  : 8;
+
+/** Versioned API-reference URLs the built site links to, grouped by line. */
+function linkedApiUrls(dir, byLine = new Map(), budget = { files: 6000 }) {
+  let entries;
+  try {
+    entries = fs.readdirSync(dir, { withFileTypes: true });
+  } catch {
+    return byLine;
+  }
+  for (const e of entries) {
+    if (budget.files <= 0) return byLine;
+    const full = path.join(dir, e.name);
+    if (e.isDirectory()) {
+      linkedApiUrls(full, byLine, budget);
+    } else if (e.name.endsWith(".html")) {
+      budget.files -= 1;
+      const html = fs.readFileSync(full, "utf8");
+      const re = /https:\/\/docs\.scandit\.com\/(\d+\.\d+)\/data-capture-sdk\/([^"'#\s<>]+)/g;
+      let m;
+      while ((m = re.exec(html))) {
+        const [, line, rest] = m;
+        if (!byLine.has(line)) byLine.set(line, new Set());
+        byLine.get(line).add(rest);
+      }
+    }
+  }
+  return byLine;
+}
+
+/** Deterministic even spread, so CI checks the same pages every run. */
+function sample(items, n) {
+  const sorted = [...items].sort();
+  if (sorted.length <= n) return sorted;
+  const step = sorted.length / n;
+  return Array.from({ length: n }, (_, i) => sorted[Math.floor(i * step)]);
+}
+
+async function status(url) {
+  try {
+    const res = await fetch(url, { method: "HEAD", redirect: "follow" });
+    return res.status;
+  } catch {
+    return 0;
+  }
+}
+
+async function fetchHtml(url) {
+  try {
+    const res = await fetch(url, { redirect: "follow" });
+    return res.ok ? await res.text() : null;
+  } catch {
+    return null;
+  }
+}
+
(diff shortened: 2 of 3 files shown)


.github/workflows/build-docs.yml
```diff
@@ -70,3 +70,27 @@ jobs:
...
+      # way at 8.6 - and then nothing links it, so the gate alone would never see
+      # the newest frozen line, the one whose content is closest to current and so
+      # the likeliest to outrank it. Discovery probes for those and prints them.
+      #
+      # `|| true` on purpose: if discovery cannot reach the site, LINES is empty
+      # and the gate runs on the linked lines instead of the step going red. The
+      # gate says in its own output when it ran without --lines.
+      - name: Verify API-reference canonical tags
+        run: |
+          LINES=$(node scripts/discover-api-reference-lines.cjs --quiet || true)
```
moritzhartmeier: **`LINES` shadows a shell special variable.**

zsh declares `LINES` as `typeset -i10`, so an assignment to it is arithmetic-evaluated:

```
$ bash -c 'LINES="8.3,8.4,8.5"; echo $LINES'
8.3,8.4,8.5
$ zsh  -c 'LINES="8.3,8.4,8.5"; echo $LINES'
8
```

CI runs `bash`, so this workflow is safe as written. The problem is the same name in the usage block the script documents for humans (`scripts/discover-api-reference-lines.cjs`, separate comment) — on macOS that runs under zsh, the value silently becomes `8`, and the gate then dies with `--lines takes major.minor values (got "8")`. I hit this verbatim reproducing the CI step locally.

Suggest renaming to `API_LINES` here and in the script header. Given that the theme of this PR is "a silent default is the failure mode", it seems worth closing.

scripts/discover-api-reference-lines.cjs
```diff
@@ -0,0 +1,381 @@
...
+ *   node scripts/discover-api-reference-lines.cjs            # human readable
+ *   node scripts/discover-api-reference-lines.cjs --quiet     # bare comma list
+ *
+ * In CI - note `node`, NOT `yarn`. This repo is on Yarn 1, which writes its own
+ * `yarn run v1.x` / `$ node …` / `Done in Xs` banner to STDOUT, so
+ * `$(yarn discover:api-reference-lines --quiet)` captures the banner instead of
+ * the comma list. The `discover:api-reference-lines` script entry is for reading
+ * the human report; anything parsing the output must call node directly.
+ *
+ *   LINES=$(node scripts/discover-api-reference-lines.cjs --quiet)
```
moritzhartmeier: Same `LINES` issue as flagged in `build-docs.yml` — and this is the copy that bites, because it is the invocation a maintainer pastes into a local shell. Under zsh `LINES=$(...)` turns `8.3,8.4,8.5` into `8`.

```suggestion
 *   API_LINES=$(node scripts/discover-api-reference-lines.cjs --quiet)
```

(and the `yarn verify:api-reference-seo ${API_LINES:+--lines "$API_LINES"}` line below it.)

scripts/verify-api-reference-seo.cjs
```diff
@@ -0,0 +1,1337 @@
...
+      (unknownLines.length || absent
+        ? ", and nothing was learned about the rest of what was asked"
+        : "") +
+      "." + "\n",
+  );
+}
+
+// Only when run directly. Without the guard, `require()`ing this file fired a
+// full live-network run.
+if (require.main === module) {
```
moritzhartmeier: **Nothing is exported, so none of this can be tested.**

This is my main ask on the PR. `attr`, `headOf`, `canonicalOf`, `isNoindex`, `hasNoindexIn`, `hasNoindexHeader` and `samePage` are pure functions, they carry most of the risk here, and nearly every one has a comment describing a bug that a two-line assertion would have caught *and kept caught*. I wrote 15 such assertions against a patched copy of this file in a couple of minutes and they all passed — they just cannot live in the repo today.

Suggest adding `module.exports = { attr, headOf, canonicalOf, isNoindex, hasNoindexIn, hasNoindexHeader, samePage }` below this guard (the guard already makes that safe), plus a `test:api-reference-seo` entry alongside the existing `test:search-facets`. The same applies to `scripts/lib/linked-api-lines.cjs`, which already exports but has no tests.

As it stands the prose is doing the job a test should do, and unlike a test it cannot fail when someone breaks the invariant it describes.

scripts/verify-api-reference-seo.cjs
```diff
@@ -0,0 +1,1337 @@
...
+    );
+    if (!currentVersion || !a.version || a.version !== currentVersion) return [];
+    return Array.isArray(a.uncertain) ? a.uncertain : [];
+  } catch {
+    return [];
+  }
+}
+
+/** The served version number, as the search-tag manifest states it. */
+function currentVersion() {
```
moritzhartmeier: `currentVersion()` here and `currentNumber()` in `discover-api-reference-lines.cjs:140` are the same function under two names, reading the same file for the same field. `ORIGIN` and `REQUEST_TIMEOUT_MS` are duplicated between the two scripts as well.

`scripts/lib/linked-api-lines.cjs` was added precisely so the two cannot drift — these belong in it too.

moritzhartmeier (commented):
I ran both scripts against a local build and the live site. They work, and the payoff is concrete: discovery found `/8.3/ /8.4/ /8.5/` — lines nothing in the build links — and the gate then confirmed all three serve duplicate content with neither canonical nor noindex. That is exactly the case the second script exists for. Cost is small too: discovery 42 requests / ~5s, gate ~4s at `--sample 4`.

I also hand-probed the parsers (`attr`, `headOf`, `canonicalOf`, `isNoindex`, `hasNoindexHeader`, `samePage`) with 15 tricky inputs — commented-out tags, `data-name` decoys, `rel="alternate canonical"`, `max-image-preview: none`, `bingbot:` scoping, protocol-relative canonicals. 14 of 15 behaved exactly as the comments say they do.

Three things below: one real bug, and two maintainability asks. Nothing I would block on.

moritzhartmeier (approved):
All three findings from the previous review are fixed, and the commits that followed closed two false-pass classes I had missed. Verified locally against a real build and the live site.

**On my findings.** `API_LINES` renamed in both the workflow and the script header. Two suites added — `test-api-reference-seo.cjs` (44 assertions, parsers + lib) and `test-api-reference-seo-verdicts.cjs` (15, verdict branches against a localhost stub) — 59 passing in ~1s, fully offline, both wired into CI. `BUILD`, `ORIGIN`, `REQUEST_TIMEOUT_MS` and `currentVersion` now live in the shared lib.

The PR-skip split is the right one: the live step is off pull requests, while the two offline test steps stay on them, since those are what actually test the code in the diff.

**Two real bugs found after my review**, both worth calling out:

1. **`servesSymbol`.** A host-level catch-all redirect made the versioned and unversioned URLs converge on `/`, `samePage` compared `"/"` against `""`, and every pick read as sound — on a line that did not exist. A false pass on the exact signal this gate verifies. Now shared so the two scripts cannot answer it differently, with three stub tests behind it.

2. **The `attr` rewrite is justified.** I ran the old regex against the new tokenizer, and the old one was genuinely exploitable:

   ```
   <link rel="stylesheet" title="href=/evil" href="/real">
   old attr(tag, "href") -> "/evil"
   ```

   `"` counted as a word boundary, so an attribute *value* could impersonate a canonical. The new tokenizer passed all 13 adversarial cases I tried: unquoted values, single quotes, spaces around `=`, self-closing tags, valueless attributes, uppercase, duplicate attributes, unterminated quotes, and the decoy above.

**Also verified.** The counterpart cache now drops non-definitive responses, so one 429 no longer poisons a shared pick across every borrowed line. `dropArtefact()` on all four early returns closes a stale-artefact path the version check cannot see, since the version does not change between builds of one release. `spreadAcrossLines` works — in my live run `/8.5/` and `/8.4/` lead the violation list instead of `/6.28/` taking all 20 slots.

**Measured live:** discovery 99 requests / 19 s (matching the figure in the comment exactly), gate 72 / 2 s — 171 requests and ~22 s total, against budgets of 250 and 260 and a 15-minute step backstop. Found `/8.3/ /8.4/ /8.5/` and 34 unsound pages.

Two notes, neither blocking:

- The full lower-major sweep costs 42 → 99 requests and finds the same three lines today, so the extra requests buy nothing at the current shape. The `/7.12/`-surviving-above-a-gap case it protects against is real and stub-tested, and the step is off PRs now, so the trade is fine — just worth knowing.
- `"... and N more, across the same lines"` holds only while the number of lines carrying violations is under the print cap of 20. Five today. Cosmetic.

My earlier point about comment volume still stands — ~4,000 lines for the feature, and the two main scripts carry a lot of first-person revision history. But with 59 tests now pinning the invariants, most of that prose is redundant rather than load-bearing, so I would treat a trimming pass as a cheap follow-up rather than a condition of merge.

Approving.

PR #433, opened 2026-08-28
docs(react-native): add AIO vs classic comparison and migration guide [SDC-33130]

konradgapinski:
Adds `docs/sdks/react-native/aio/vs-classic.md` — a comparison of the React Native
AIO components against the classic mode-plus-view API, plus a migration section.

Jira: [SDC-33130](https://scandit.atlassian.net/browse/SDC-33130)

## What's in it

- **Side-by-side comparison** of each AIO view against its classic equivalent —
  what you write, what the AIO view does for you, and where the prop surfaces differ.
- **Migration walkthrough** for two modes, BarcodeCapture and LabelCapture, plus a
  short paragraph pointing at the rest. Two worked examples rather than one per
  mode, deliberately: the mechanical steps repeat and a reader only needs the shape.
- **Sidebar entry** under the `AIO Components` category.

## How the snippets were sourced

Every code sample is lifted from sample code that runs in-repo — `DebugApp/app/samples/<Name>/`
and `DebugApp/app/ar/barcode-ar.tsx` in `data-capture-sdk` — not written from API
signatures. That matters here because several of the call sequences are not
guessable from the types: `BarcodeArView` needs an explicit `view.start()` or it
renders a preview that never scans, `BarcodeCountView` takes its listeners through
the ref rather than as props, and `IdCaptureAioView` uses `didCaptureId`/`didRejectId`
rather than `didScan`.

## Merge conflict to expect

`sidebars.ts` will conflict with #432 — both branches add the same `AIO Components`
category at the same line. Resolve by unioning the two `items` arrays; neither side
is wrong.

## Verification

- `node scripts/docs-gate/index.cjs` — 1 doc checked, 0 errors, 0 warnings (frontmatter, links, cspell, Vale).
- `npx docusaurus build` — clean. `onBrokenLinks` and `onBrokenAnchors` are both `throw`,
  so a green build is also the link and anchor check.
- Page rendered and read locally at `/next/sdks/react-native/aio/vs-classic/`.

### DIFF
--- a/docs/sdks/react-native/aio/vs-classic.md
+++ b/docs/sdks/react-native/aio/vs-classic.md
@@ -15,7 +15,7 @@ Version 8.7 adds a new, simplified way to build a scanning screen in React Nativ
 
 The AIO components internalize much of the complex logic in the classic components, leaving you to focus on your scanning workflow rather than on the component lifecycle.
 
-Nothing was taken away. Every classic API still works exactly as it did, and existing screens keep running untouched. This page explains what the AIO components do for you, which API to reach for, and how to migrate a screen.
+Every classic API still works exactly as it did, and existing screens keep running untouched. This page explains what the AIO components do for you, which API to reach for, and how to migrate a screen.
 
 <!-- TODO(SDC-33129): link the getting-started guide once it is published. The page is
      docs/sdks/react-native/aio/get-started.md, so the link becomes:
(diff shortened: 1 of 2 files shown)


docs/sdks/react-native/aio/vs-classic.md
```diff
@@ -0,0 +1,397 @@
...
+| --- | --- | --- |
+| [Barcode Capture](/sdks/react-native/barcode-capture/get-started) | mode + `DataCaptureView` | `BarcodeCaptureAioView` |
+| [Barcode Batch](/sdks/react-native/matrixscan/intro) | mode + `DataCaptureView` | `BarcodeBatchAioView` |
+| [ID Capture](/sdks/react-native/id-capture/intro) | mode + `DataCaptureView` | `IdCaptureAioView` |
+| [Label Capture](/sdks/react-native/label-capture/intro) | mode + `DataCaptureView` | `LabelCaptureAioView` |
+| [SparkScan](/sdks/react-native/sparkscan/intro) | `SparkScanView` | `SparkScanAioView` |
+| [Barcode Count](/sdks/react-native/matrixscan-count/intro) | `BarcodeCountView` | `BarcodeCountAioView` |
+| [Barcode AR](/sdks/react-native/matrixscan-ar/intro) | `BarcodeArView` | `BarcodeArAioView` |
+
+<!-- TODO(SDC-33130): link each AIO component in the third column to its API
```
ilber: @konradgapinski is this TODO still valid? Can we do something about it?

